PR #2, opened 2017-12-26
Travis automation

jason-p-pickering:
All builds of master will be auto-deployed. Stable branches will only deploy when tagged.


jason-p-pickering:
@kjesvale I am going ahead and merging this pull request, which will automate the build and deploy of this app with Travis.

PR #3, opened 2018-02-05
Add reducer tests, fix minor reducer bugs

kjesvale:
* Added tests for job- and message reducers
* Added moduleNameMapper for jest in config
* Added a mock for i18next, might want to move this out of messageReducer test later
* Changed message reducer to use an ID counter instead of timestamps for triggering rerendering
    - Timestamps weren't needed


jenniferarnesen:
Tests look good. One convention that I think is pretty standard is to write the expect like the following, if you were to put it in a sentence:

"Expect the result to be what I expected."

In other words, this:  expect(actualState).toEqual(expectedState);
instead of this:  expect(expectedState).toEqual(actualState);

PR #14, opened 2019-02-13
Update headerbar

Birkbjo:
Removing the old d2-header-bar also removed the addition of the used rxjs-prototypes.
So I upgraded to rxjs@6 and the necessary other dependencies(redux-observable@1, redux@4, react-redux@6).

Also upgraded webpack@4 to minimize bundling with rxjs.


app/src/actions/epics.js
```diff
@@ -1,101 +1,123 @@
 import * as actions from 'constants/actions';
-import { combineEpics } from 'redux-observable';
-
+import { combineEpics, ofType } from 'redux-observable';
+import { concatMap, switchMap } from 'rxjs/operators';
```
kjesvale: 🧐😄

PR #15, opened 2019-02-13
fix: Do not use schemas, as jobConfiguration is not metadata

Birkbjo:


Birkbjo:
jobConfigurations was reverted to be metadata once again, since it broke the importer. 
Ref: https://github.com/dhis2/dhis2-core/pull/2902

PR #23, opened 2019-06-21
refactor: port scheduler app to app platform

ismay:
Jira epic for this PR: https://jira.dhis2.org/browse/TECH-214


ismay:
## Useful links

- https://sketch.cloud/s/DV15W/a/plkG4k/play
- https://github.com/dhis2/design-system
- https://github.com/dhis2/ui-core
- https://github.com/dhis2/usage-analytics-app
- https://github.com/dhis2/reports-app
- https://medium.com/fed-or-dead/handling-spacing-in-a-ui-component-library-70f3b22ec89
- https://material-ui.com/system/spacing

## Todo

#### Initial tasks:

- [x] Enable cli-style (also on travis)
- [x] Set up eslint and stylelint
- [x] Add favicons and other static files
- [x] Use the app-runtime and new headerbar (from ui-widgets)
- [x] Add designs to this issue for reference
- [x] Import changes from unejected apps (see the PRs)
- [x] Plan out refactor

#### Styling:

- [ ] Import and set roboto
- [ ] Check for any other theme vars that need to be set
- [ ] Import theme from ui-core
- [ ] Need components for (check css approach, sc?):
  - [ ] horizontal arranging (Arrange? Flex?)
  - [ ] padding & margin (Box)
  - [ ] pagewrapper
  - [ ] title

#### Experimental:

- [ ] Test netlify deployments with play backend
- [ ] Code coverage reporting
- [ ] Bundle size analysis
- [ ] Run stylelint on commit as well, to catch errors before pushing
- [ ] Prettify css on commit as well

#### Technical:

- [x] Move to css-modules for styling
- [ ] Move to thunks for actions
- [ ] Replace custom components with ui-core / ui-widgets components
- [ ] Use final-form for forms, in preparation for ui-forms
- [ ] Proper tests
- [ ] Prop-types
- [ ] Ensure translations are working correctly
- [ ] Verify nothing from #21 got lost

#### UI-core

- [ ] Maybe a top-level container for wrapping apps would be nice (bg color and padding)
- [ ] Abstract useful components to ui-core
- [ ] Load only the fonts that we need, instead of importing typeface-roboto? (system fonts?)
- [ ] Sizing in px?

#### After refactor:

- [ ] Add fossa badges
- [ ] Add deploy build script to travis
- [ ] Test css variables component (untestable atm.: https://github.com/jsdom/jsdom/issues/1895)?

ismay:
> Hi! Due to a lack of activity on this issue over time (7.776*10^9 ms and counting, to be precise) it seems to be stale. If there is no further progress on it, it will be closed automatically.
> 
> If this is still relevant, maybe there is something you can do to move it forward? For example provide further information in a comment, or supply a PR? Any activity on this issue will keep it open. Thanks! 🤖

Still relevant, waiting for required components in ui-core.

Mohammer5 (approved):
The app works quite well!

There are just a couple of minor things, that don't need to be addressed now:

### 1. Sometimes there are unexpected error message that I can't get a lot of info out of

```
Something went wrong whilst creating your job

    Failed to validate job runtime - `Remote server is not configured`
```

### 2. Input validation could be refined

When adding/editing a job, some fields expect a number that "expect" either 0 or a positive integer, but I can submit the form with negative values. I think we should add a validator for this, I have no idea if a negative value could cause harm on the server. I assume some would need 0 to Infinity and some others 1 to Infinity, not sure though

ismay:
Thanks for the review!

> There are just a couple of minor things, that don't need to be addressed now:
> ### 1. Sometimes there are unexpected error message that I can't get a lot of info out of
> 
> This one, for example, happened, when I tried to add a data synchronization job:
> 
> ```
> Something went wrong whilst creating your job
> 
>     Failed to validate job runtime - `Remote server is not configured`
> ```
> 
> I guess there's not much you can do about that.. And I also know that error handling is a bit weird and inconsistent.

Yeah I believe that that's because there needs to be a bit of configuration that has to be set up on the server, besides just the job configuration. But yeah, completely agree, this error message is only helpful if you already know how job configuration works exactly. Other submission errors could be improved as well I think, it's fairly inconsistent for the job scheduler endpoints. I'll create an issue for it, it seems like after release it would be good to take a look at the user friendliness and consistency of the errors the backend returns.

> ### 2. Input validation could be refined
> 
> When adding/editing a job, some fields expect a number that "expect" either 0 or a positive integer, but I can submit the form with negative values. I think we should add a validator for this, I have no idea if a negative value could cause harm on the server. I assume some would need 0 to Infinity and some others 1 to Infinity, not sure though

Yeah this is a bit tricky. For example, Scott opened an issue that explains that for monitoring and predictor jobs, the number inputs actually can also accept negative values: https://jira.dhis2.org/browse/DHIS2-10466. I don't think that there's currently a way to deduce what the fields accept from the backend job definitions I'm getting. All I'm getting is the type of value that's expected (number, string, etc.), but not the range.

My suggestion to solve it is that after release we move the app to static forms (like the usage-analytics-app), https://jira.dhis2.org/browse/DHIS2-10491. That way we can also implement Scott's suggestion of better helptexts with the fields, plus we can also add better client-side field validation (alongside also just a nice reduction in the complexity of the app).

Mohammer5:
Sounds good to me. I don't see a way to work around those issues with the current semi-dynamic strategy

Mohammer5 (approved):

ismay:
> Sounds good to me. I don't see a way to work around those issues with the current semi-dynamic strategy

Yeah exactly, me neither.

HendrikThePendric (approved):
Code is looking super clean 👍

PR #24, opened 2019-09-17
fix: add translations for new job types

ismay:
This adds two new (english) translations for the job types. Do not merge unless we have an OK from QA!


ismay:
> Question: Why are we not using the standard d2-i18n implementation?

No idea. I would also expect it to be implemented similarly to our other apps, but for some reason it isn't.

Mohammer5:
> No idea. I would also expect it to be implemented similarly to our other apps, but for some reason it isn't.

I think we should add that to the epic

ismay:
> I think we should add that to the epic

Good idea. I've updated the epic. There was already something about adding translations, but I've updated it to also state that it should be a dhis2 standard implementation.

PR #36, opened 2020-02-20
chore: update to ui-forms and ui-core@latest

ismay:
Picking up the refactor for the scheduler app. I'll submit separate PRs per topic, instead of piling onto the sizeable number of commits on #23.

I'm afraid that the changes are less atomic than I'd like. But ui-core was quite far behind. And instead of updating ui-core first, and then adding ui-forms, I thought I'd skip the inbetween step to save on time. It makes for a larger PR but it's a bit quicker.

Also, I'm aware that a lot of best practices have emerged in our ecosystem since I last worked on this codebase. Don't worry, I aim to align it fully with that, it's just that I'd prefer to do that in separate PRs for reviewability.


ismay:
No rush btw @Mohammer5, just tagged you for later so I don't forget.

ismay:
> This looks good to me.
> I think we can import `Field` and `FormSpy` from `ui-forms` and remove `react-final-forms` as an explicit dependency for this app.

Yeah you're right, good catch. I've talked to Hendrik, and later on we could potentially also reexport from final-form, so that that can be removed as well. For now, I've removed the react-final-form imports in favor of those from ui-forms.

PR #37, opened 2020-02-27
refactor: update job list route

ismay:
This update the JobList route to the specs stated in the design (https://sketch.cloud/s/DV15W/a/7knaz7). Certain things are still missing (see bottom todo list), I'll implement those in later PRs. I'm planning to tackle the `<Tag>` component for ui-core first, as we'll need that for this app as well.

Todo:

- [x] Revisit LinkButton
- [x] Remove unused components
- [x] Proper values for JobList header layout (see design)
- [x] Title and `i` icon styling
- [x] Job status has to be displayed in a Tag component
- [x] Add/update tests

Later PRs:

- [ ] Table should be sortable
- [ ] Strings should be translated
- [ ] Disabled jobs should be styled differently
- [ ] Add tests for FormFields, Form and Pages
- [ ] Verify job status enum, couldn't find it in the docs, so took it from dhis2-core source


awgaan:
Looks good.

## Questions
1. In `JobList`: why are you using `history.push` directly instead of wrapping the `Button` in a `Link`? Easier to stub out `history.push` I guess, but it feels like we should use the navigation functionality provided by `react-router-dom`, or?
2. i18n: are you leaving this for later?
3: Button disabled while fetching: why remove this functionality?

## Thoughts
I like the way you're using `CSSVariables` to follow the design specifications. Not sure if this is the norm, but I like the simplicity and elegance of this approach.

ismay:
Thanks for the review! I'll address your points below:

> In `JobList`: why are you using `history.push` directly instead of wrapping the `Button` in a `Link`? Easier to stub out `history.push` I guess, but it feels like we should use the navigation functionality provided by `react-router-dom`, or?

Yeah I would prefer to leverage react-router as well. So the design calls for a button, but nesting a button in a link would not be valid html5 unfortunately. Another solution (which I used before), would be to style a react router link to look like a button.

But I felt that that was a whole lot of code and maintenance. Using an `<a>` is technically correct in my opinion, but weighed against the simplicity of just using a button with history.push I think that just the regular button seems more practical.

>     2. i18n: are you leaving this for later?

Yeah. There are a couple of things that I want to address later (all the points under `Later PRs` in the top comment). I think they're big enough to warrant their own PR.

>        3: Button disabled while fetching: why remove this functionality?

Yeah so it did seem the most safe to me to disable the ui whilst fetching. As you could otherwise trigger an action whilst another action is still fetching, potentially screwing up the state (it did look a bit janky ui-wise though).

But I figured that in moving all requests to the app-platform's data-query there's no more central state anyway, so disabling based on central state won't be possible anymore either. So I thought I'd simplify things and remove those checks.

> I like the way you're using `CSSVariables` to follow the design specifications. Not sure if this is the norm, but I like the simplicity and elegance of this approach.

Yeah the idea is that for apps that use vanilla css this component can be used to import our theme variables. With styled-jsx you can import the theme directly in js of course, but since we don't use that here this seemed like a nice way to do it.

Let me know what you think!

ismay:
By the way, would you mind approving this PR if you think it's good to go @awgaan? Or otherwise, let me know which points you'd like to address. Thanks!

ismay:
> Oh. Sorry. I thought I already did that with the previous comment.
> My bad.

N.p., thanks!

PR #38, opened 2020-03-03
refactor: update to dhis2 app-platform

ismay:
This updates the dhis2 related libs that weren't up to date. So `app-runtime`, `ui-core`, `cli-app-scripts`, `cli-style`. These updates also mean that state management was changed from redux to using the data-engine, which has implications throughout the app, so this PR has quite some changes. Currently there is local state in some components (useState), but no single globally shared store.

That means that currently fetched data doesn't persist in the app (though the browser might cache it, etc.). Eventually the app-runtime should handle caching so initially I decided to build with that endgoal in mind. It does mean that some parts of the ui are more janky than they'd be with caching, so eventually we might have to have an intermediate solution that adds some persistence. I would like to keep that completely separate though, so that once caching is in the app-runtime, those additions can be removed easily.

For certain other parts of the app we might still need global state, but I will also tackle that in a different PR. Before merging I think squashing all commits to: `refactor: update to dhis2 app-platform` makes the most sense.

Todo:

- [x] Use moment instead of date-fns (added a comment where moment is being used, the `nextExecutionTime` we receive is not always correct, needs to be addressed on the backend)

Later, separate PRs:

- RunJobModal uses a non-standard way to run jobs, because the endpoint is slightly unusual. See if this can be addressed better later on. Same goes for the parameter options hooks (which allow for dynamic queries, something the data engine does not allow currently).
- backend: nextExecution time job property should never be in the past (now it sometimes is)
- Address console errors caused by RFF (or document)
- Fetched data persistence / global app state (if necessary)
- Update add job styling to design specs

Ref: #23


d2.config.js
```diff
@@ -0,0 +1,8 @@
+const config = {
+    type: 'app',
+    entryPoints: {
+        app: './src/components/App',
+    },
```
awgaan: `coreApp: true`
needs to be set for all apps that are bundled into the main DHIS2 WAR file.
`title` and `name` should also be set.

ismay: Ah I didn't know that, thanks. Updated it in 19cdb39

ismay:
> Excellent use of the `app-runtime`.
> In particular I think `useGetLabeledOptions` handles dynamic queries nicely without too much added complexity.
> 
> https://github.com/dhis2/scheduler-app/blob/df0a44bd876bf6f0af33b3356b397be24d8cf1b1/src/hooks/parameter-options/use-get-labeled-options.js#L4-L57

Thanks, yeah I like how this hook does not bother the components too much with the exception for dynamic queries. If the app-runtime starts allowing dynamic queries or another kind of solution we should be able to just update this hook.

I do think that it duplicates a fair bit of logic that's probably already in the data-engine as well, but at least it's a stable solution for now.

PR #39, opened 2020-03-25
refactor: fix next run time display

ismay:
Small inbetween fix following up on a todo point from #38. I thought that there was a bug in the jobConfigurations endpoint, as it seemed to be returning timestamps in the past.

But after talking to Morten about it I understand that this endpoint can be assumed to return UTC time (see slack [here](https://dhis2.slack.com/archives/C1XNHEY0L/p1585128053002800)). So I'm now parsing it as UTC and have abstracted it to a separate component and added some tests.

### DIFF
--- a/src/pages/JobList/JobNextRun.js
+++ b/src/pages/JobList/JobNextRun.js
@@ -1,0 +1,37 @@
+import moment from 'moment'
+import { string, bool } from 'prop-types'
+
+const JobNextRun = ({ nextExecutionTime, enabled }) => {
+    if (!enabled) {
+        return '-'
+    }
+
+    const now = moment(Date.now())
+
+    /**
+     * The recommendation is to run dhis2 on a server set to UTC time.
+     * In that case this timestamp is also UTC. If those recommendations
+     * weren't followed the time could be off, but there's nothing
+     * we can do to detect that.
+     */
+    const nextRun = moment.utc(nextExecutionTime)
+    const nextRunIsInPast = nextRun.isSameOrBefore(now, 'minute')
+
+    /**
+     * If the time is in the past, that could mean that the task is running,
+     * and the nextExecutionTime hasn't been updated yet.
+     */
+
+    if (nextRunIsInPast) {
+        return 'Not scheduled'
+    }
+
+    return now.to(nextRun)
+}
+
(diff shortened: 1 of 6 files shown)


ismay:
Updated the string displayed under 'next run' for disabled jobs to match the design (see here: https://sketch.cloud/s/DV15W/a/plkG4k/play).

src/pages/JobList/JobNextRun.js
```diff
@@ -0,0 +1,43 @@
...
+        return 'Not scheduled'
+    }
+
+    /**
+     * According to the moment docs you would expect to() to be the
+     * correct method, as all those examples are describing future events.
+     * But using from() actually results in future relative time,
+     * so that is assumed to be a mistake in the moment docs.
+     */
+    return nextRun.from(now)
```
awgaan: I think this should be `now.to(nextRun)` using `now` as the base date.

ismay: Ah yes of course. That makes more sense, thx 👍. Fixed in 48f5160

ismay:
> Looks good to me.
> 
> `* The recommendation is to run dhis2 on a server set to UTC time.`
> I could not find this recommendation anywhere in the documentation.

Hmm ok. Yeah I took that from my conversation with Morten on slack (see the link above), I thought it was a general recommendation.

So it's tricky. If I understand correctly, since dhis2 doesn't store TZ information, nor send TZ information to the client, there's no way a client can know what TZ the timestamp should be parsed as.

My initial approach was to parse as local time. But since debug was running on UTC that didn't work for example. So from my conversation with Morten's I understood that he'd recommend the server run on UTC. I thought that that would also be recommended in the setup docs, but checking out the docs yields this:

Screenshot 2020-03-26 at 09 53 40

> From: https://docs.dhis2.org/master/en/implementer/html/install_server_setup.html

So we actually recommend setting the server TZ to the TZ the server is located in. So now I'm confused. The fix in this PR would work for servers set to UTC, but it wouldn't work for servers set to local time. If we recommend setting servers to local time, then I don't understand why our own servers are set to UTC. We'll probably need to check with the others and see which direction to take, I'll ask on slack.

PR #42, opened 2020-04-02
refactor: use dhis2 prop-types

ismay:
Use our own prop-types package.

Ref: https://github.com/dhis2/prop-types/pull/125 & https://github.com/dhis2/prop-types/pull/127


ismay:
> Approved. Probably a good idea for all of our apps to use `@dhis2/prop-types` instead of `prop-types`.

Yeah I agree. If I'm not mistaken that is also the eventual goal (for our libs as well).

PR #45, opened 2020-04-27
chore: update eslint config

ismay:
Enabling some useful rules, see: https://github.com/benmosher/eslint-plugin-import. This adds checks for common import mistakes/anti-patterns (like no cyclical dependencies, no unused modules, etc.). Other than that it enables some style rules that allow us to have to think less about how we're writing our imports, since `eslint --fix` will keep it consistent for us.

* The `import/no-internal-modules` rule enforces the existing repo structure. It ensures that the index.js of every folder in `components`, `hooks`, `pages` and `services` is always the external boundary. Which will hopefully make the interaction between the different parts of the app easier to reason about, we'll see how it works out in practice though.
* I've kept the plugin's bundled warnings and errors. We're setting most of them manually, but I figured it couldn't hurt to have their settings as a base/default.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -4,5 +4,6 @@
 coverage
 node_modules
 public
+src/locales
 
 d2.config.js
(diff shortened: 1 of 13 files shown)


ismay:
> Those rules look helpful. Approved.

Thanks!

PR #46, opened 2020-04-27
fix: revert to old translation strategy

ismay:
The translations in this app were broken in a couple of places (and I noticed my fix from earlier was incomplete). I tracked it down to [this line](https://github.com/dhis2/scheduler-app/blob/ab3cd673111847520d5464072cb04dafda3053d8/src/utils/configI18n.js#L19), where the namespace used is no longer correct (but `default` works as expected).

See [this issue](https://github.com/dhis2/app-platform/issues/369) for more info on why using the `default` namespace fixes the translations. This fix means that the translations in [apiTranslations.js](https://github.com/dhis2/scheduler-app/blob/ab3cd673111847520d5464072cb04dafda3053d8/src/utils/apiTranslations.js) are now picked up again.

Todo:

- [x] Double check all translations and hardcoded strings (there were a couple of duplicates, etc.)
- [x] Do we still need the map in redux from uppercased jobtype name to translated name?

Ref: https://jira.dhis2.org/browse/DHIS2-8709
Docs: https://www.i18next.com/principles/namespaces

As for the commits, I propose we squash them to the PR title before rebasing on master.

### DIFF
--- a/i18n/en.pot
+++ b/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
-"POT-Creation-Date: 2020-01-29T15:16:04.305Z\n"
-"PO-Revision-Date: 2020-01-29T15:16:04.305Z\n"
+"POT-Creation-Date: 2020-04-28T12:20:25.090Z\n"
+"PO-Revision-Date: 2020-04-28T12:20:25.090Z\n"
 
 msgid "Scheduler"
 msgstr ""
@@ -27,6 +27,9 @@ msgid "Are you sure you want to delete this job?"
 msgstr ""
 
 msgid "Add new job"
+msgstr ""
+
+msgid "Add job"
 msgstr ""
 
 msgid "Required"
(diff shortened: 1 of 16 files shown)


i18n/en.pot
```diff
@@ -154,21 +157,3 @@ msgstr ""
...
-msgstr ""
-
-msgid "Every day at three am"
-msgstr ""
-
-msgid "Every weekday at noon"
-msgstr ""
-
-msgid "Every week"
-msgstr ""
```
amcgee, line 174: Question - where did these strings come from and where did they go?  @ismay

ismay: Hey, so these were part of an incomplete earlier fix. They're labels for the cron expression presets. The translations can also be found here:

https://github.com/dhis2/scheduler-app/blob/626e4ea32d12393941ff5a8f23ea1a83b29fc3c8/src/utils/apiTranslations.js#L10-L16

Those are the translations that this PR reverted back to, so the above (removed) translations could be safely removed.

amcgee: Ok, but does that mean these strings are only available in English and Urdu?  Is there a mechanism for hooking this up to Transifex?  Sorry if I'm missing some context here, not sure what mechanism of i18n this is using, want to make sure we have a full picture before the overhaul

ismay: > Ok, but does that mean these strings are only available in English and Urdu?

That's right. At the moment there's only Urdu and English

>  Is there a mechanism for hooking this up to Transifex?

Don't know. This app hasn't been hooked up to transifex yet. Maybe we'll have to make some change before hooking it up to transifex.

> Sorry if I'm missing some context here, not sure what mechanism of i18n this is using, want to make sure we have a full picture before the overhaul

So what this PR does is just fix something that broke in the app's i18n setup. This isn't the refactored version, so I wouldn't pay too much attention to this version of the app. It'll be replaced soon anyway.

amcgee: Got it, thanks @ismay!

I'm going to make a note of it in dhis2/app-platform#369 so we don't forget to migrate it to the new system, since it might be easy to miss this non-standard i18n approach

ismay: Thanks. Yeah it also caught us by surprise.

PR #49, opened 2020-04-30
fix(prop-types): update prop types

ismay:
Updates to the latest release of our prop-types lib.


ismay:
> Looks good.
> Though I wish we could de-structure right in the import statement (I know ES2015 names imports does not support this).

Yeah me too, it feels a bit clunky this way.

PR #52, opened 2020-06-22
ci(travis): fix build error

ismay:
Fixes the buid failure on the refactor-scheduler branch. For some reason upgrades to our libs sometimes require you to `rm -rf node_modules && yarn`. This is a temporary fix to address that, though in the long run we should definitely make sure that our libs don't require this, because it's affecting the reliability/reproducibility of our dependency installations.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+cache:
+    npm: false
 language: node_js
 node_js:
     - lts/*


awgaan:
I met this error while working on the `import-export-app` as well. There it helped to manually clear the build cache. All subsequent builds were successful. Did you try that here?

ismay:
> I met this error while working on the `import-export-app` as well. There it helped to manually clear the build cache. All subsequent builds were successful. Did you try that here?

Yeah I did clear the cache, but somehow travis was persisting it for subsequent builds.

ismay:
Thanks!

PR #53, opened 2020-06-29
feat: add edit route

ismay:
See design: https://www.sketch.com/s/b5ff707f-5900-4e58-baf0-71cc599dc295/a/0OGn5V

I've also improved the test coverage, since the app is now pretty close to its final functionality. I've ignored certain areas with istanbul ignore, so that the _reported_ test coverage is now 100%. I've done that because that way it's easy to see which areas we've unintentionally not tested yet (as they'll be the only ones not at 100% in the report). Plus, it's apparent in the source which areas are potentially untested and it requires conscious effort to add an istanbul ignore comment (which will then have an attached git blame as well).

Test approach has also been slightly revised to make it more uniform across all the tests. I've also removed snapshot tests as they're easy to accidentally update, in which case they're no longer useful.

In my opinion there's definitely room for improvement with the tests. Sometimes they're a bit too close to testing implementation details. And I think that instead of mounting at times it'd be better to use shallow with [.dive](https://enzymejs.github.io/enzyme/docs/api/ShallowWrapper/dive.html) & [.shallow](https://enzymejs.github.io/enzyme/docs/api/ShallowWrapper/shallow.html), that way we could mock less.

But I'll address all of that, plus cleaning up the components in `/pages` and smoother loading of fields (sometimes a bit jerky now in places) in later PRs.


src/components/FormFields/LabeledOptionsField.js
```diff
@@ -2,20 +2,35 @@ import React from 'react'
...
 const LabeledOptionsField = ({ endpoint, label, name, parameterName }) => {
-    const { loading, error, data } = useGetLabeledOptions({
-        endpoint,
-        parameterName,
+    /**
+     * HACK: this is a bit of a hack to allow using the useDataQuery hook with
+     * a dynamic query. Initially we used a custom hook for this but that
+     * replicated all of the internal logic of the useDataQuery hook so this
+     * seems like a better trade-off.
+     */
```
awgaan, line 31: I 100% understand why we ultimately ended up with this solution, but it's not optimal, is it?
I feel like we meet this issue in every app-platform app. The `app-runtime` must provide a better solution, otherwise the whole benefit of static declarative queries is gone. Or am I wrong?

src/components/Modal/DeleteJobModal.js
```diff
@@ -8,11 +8,17 @@ import {
...
+
+const mutation = {
+    resource: 'jobConfigurations',
+    id: /* istanbul ignore next */ ({ id }) => id,
+    type: 'delete',
+}
 
 const DeleteJobModal = ({ id, hideModal }) => {
-    const [deleteJob] = useDeleteJob()
+    const [deleteJob] = useDataMutation(mutation)
```
awgaan, line 21: Maybe also extract out `error` and `loading`, so that we can show errors or disable the primary action button while we're loading. This comment is true for many files. Not sure what your approach to error handling is for this app, or if you intentionally left that for later.

awgaan:
> In my opinion there's definitely room for improvement with the tests. Sometimes they're a bit too close to testing implementation details. And I think that instead of mounting at times it'd be better to use shallow with .dive & .shallow, that way we could mock less.

I think the tests are completely fine. It seems like it's very easy to unintentionally write tests that test implementation details rather than behavior when using `enzyme`. But I don't think the effort of switching to another **testing library** is worth it.

src/components/FormFields/LabeledOptionsField.js
ismay: I agree. I also think that it would be best to address this in the app-runtime. I think a dynamic query is common enough to warrant a solution. Maybe I'm not aware of an existing solution. At any rate, this is definitely not ideal.

src/components/Modal/DeleteJobModal.js
ismay: Yeah you're right, and it's true that it happens in several areas. I do still have to take care of that. Partially I'm also thinking it'd be good to talk about those areas of the app with Joe. Every area of the app should show feedback whenever it's doing something (like fetching, etc.), or has an error.

ismay: I'm thinking to care of this separately (separate from this PR). What do you think?

ismay:
> > In my opinion there's definitely room for improvement with the tests. Sometimes they're a bit too close to testing implementation details. And I think that instead of mounting at times it'd be better to use shallow with .dive & .shallow, that way we could mock less.
> 
> I think the tests are completely fine. It seems like it's very easy to unintentionally write tests that test implementation details rather than behavior when using `enzyme`. But I don't think the effort of switching to another **testing library** is worth it.

Yeah. I mean, they're a good enough base. But I would like to avoid mocking any internals (like the data fetching), and ideally only use selectors that mimick how the user traverses the app.

I know that react-testing-library makes a big deal of enforcing proper testing by encouraging the above with certain selectors etc. I agree with the goal, but I'm not too sure if I agree with their solution. I've tried it for a bit with this app, but it seemed to me like a less mature library than enzyme. Incorrect documentation etc.

Throwing away enzyme and starting from scratch seems like it's a bit wasteful for criticism that seems to only affect a small surface area of enzyme. Maybe I'm mistaken, but I've become rather skeptical whenever a new library gets popular based on some flaw of an older library and then starts completely from scratch, stating that it'll solve all our problems. Why not contribute back to enzyme and fix it there? Or maybe create an eslint preset that enforces it.

Anyway, I guess what I'm trying to say is that:

- At the moment, to me enzyme seemed like the most mature solution (but I've not done a _lot_ with react-testing-library, so I could be wrong), but the selectors I'm using should mimick user behaviour more closely (so select by text content, etc.)
- I would like to avoid mocking internals, like the data-fetching. I've experimented with [nock](https://github.com/nock/nock) a bit, but I need to be able to await a promise in the test that indicates that the async logic in the component is done. I could not find an elegant way to do that with some of the components in this app. With redux the separation of actions and components made this a bit easier, but with our current setup I'm not sure how best to achieve this.

Hope that that makes sense. Let me know what you think.

ismay:
By the way, I'm mostly trying to outline my thought/decision process in the comment above. I agree with what you're saying about switching testing libs not being worth it.

src/components/Modal/DeleteJobModal.js
awgaan: That sounds like a good idea. Separate PR seems appropriate.

awgaan:
> Why not contribute back to enzyme and fix it there?

Then you miss out on the profits made from selling courses teaching the library you wrote. :smile: 

> At the moment, to me enzyme seemed like the most mature solution

Agreed.

> but I need to be able to await a promise in the test that indicates that the async logic in the component is done.

Could [`enzyme-wait`](https://github.com/etiennedi/enzyme-wait) help here, assuming the completion of the async logic has an immediate effect on the DOM?

ismay:
> > Why not contribute back to enzyme and fix it there?
> 
> Then you miss out on the profits made from selling courses teaching the library you wrote. 😄

Haha 😅, yeah I think you have a point there.

> > but I need to be able to await a promise in the test that indicates that the async logic in the component is done.
> 
> Could [`enzyme-wait`](https://github.com/etiennedi/enzyme-wait) help here, assuming the completion of the async logic has an immediate effect on the DOM?

Ah! That could be the solution. I'll try it. Thanks!

ismay:
All right, merged. Thanks for the review and the feedback @awgaan, appreciate it 👍!

ismay:
I will try and create a list of the remaining tasks and update the related Jira issue. Should give us a clear picture of everything that's left to do.

PR #60, opened 2021-01-11
fix(locale): use i18n for locale

ismay:
This uses i18n to retrieve locale. That way components don't have to all individually fetch it.


src/components/Cron/HumanReadableCron.js
```diff
@@ -1,11 +1,8 @@
 import { PropTypes } from '@dhis2/prop-types'
 import { useHumanReadableCron } from '../../hooks/human-readable-cron'
 
-const HumanReadableCron = ({ cronExpression }) => {
-    const humanReadableCron = useHumanReadableCron(cronExpression)
-
-    return humanReadableCron
-}
+const HumanReadableCron = ({ cronExpression }) =>
+    useHumanReadableCron(cronExpression)
```
HendrikThePendric, line 5: Having a component that only returns the return value of a hook seems a bit over-engineered. I guess this is not really related to the current PR, and there could be valid reasons for implementing it like this.

src/hooks/locale/locale.js
```diff
@@ -0,0 +1,27 @@
+import { useDataQuery } from '@dhis2/app-runtime'
+import { getLocale } from './selectors'
+
+const fallback = 'en'
+
+const query = {
+    userSettings: {
+        resource: 'userSettings',
+    },
+}
```
HendrikThePendric, line 10: A GET to `userSettings/keyUiLocale` would be more clean.

src/components/App/App.js
```diff
@@ -1,16 +1,22 @@
...
+    const locale = useLocale()
+
+    return (
+        <LocaleContext.Provider value={locale}>
+            <CssVariables spacers colors />
+            <PageWrapper>
+                <Routes />
+            </PageWrapper>
+        </LocaleContext.Provider>
+    )
```
HendrikThePendric, line 19: I'm not sure that unconditionally rendering the app is a good thing here. Effectively this is first going to render the app with an empty locale string first and then re-render the app with the correct locale. IMO it'd be more appropriate to prevent the app from rendering until the local has been set correctly.

src/hooks/locale/locale.js
```diff
@@ -0,0 +1,27 @@
...
+    if (loading) {
+        return ''
+    }
+
+    // Fall back to 'en' in case of errors
+    if (error) {
+        return fallback
+    }
+
+    return getLocale(data.userSettings) || fallback
```
HendrikThePendric, line 24: This ties in with the comment I made about `src/components/App/App.js`:
- Not sure if falling back to `en` on an error is such a great idea. Perhaps it does make sense...
- Maybe instead of returning a string, this hook could return an object similar to `useDataQuery` with `loading` `error` and `locale` (string).

HendrikThePendric (changes requested):
I've left a few comments. But in general terms I think the problem you are solving should have been solved for you by the app-platform. We should introduce a new `current-user-service` which an app can just tap into to synchronously read these type of things. 
CC: @amcgee 

Edit: Just to clarify the review status, I've left 4 comments and this one https://github.com/dhis2/scheduler-app/pull/60#discussion_r555665671 is the one I regard as a possibly causing real issues. The other ones aren't really a big deal.

src/components/Cron/HumanReadableCron.js
ismay: Yeah it is. Reason it's currently like this is because if I'd be using the hook I would have to call it conditionally, and I believe that's a no no for hooks. There's probably a better way to solve it, but that's the reason it's like this for now.

src/hooks/locale/locale.js
ismay: Ah, yeah that sounds good, I'll change it

src/components/App/App.js
ismay: Yeah, all the components that use locale can cope with an absence of the locale though. And there are just two locations where it's being used. I prefer rendering something while data is being fetched over blocking the ui from rendering while data is being fetched.

src/hooks/locale/locale.js
ismay: Yeah in that case I could just use the data query hook directly. The reason I'm doing this is because we're going to have to render something if there's no locale. English seemed like a reasonable fallback to me.

ismay: By the way, I should note that this locale is only used to localize the cron translation. And that defaults to english. So what I'm doing here just makes the existing default explicit.

HendrikThePendric: Aha, OK makes sense

HendrikThePendric (approved):
Thanks for replying to my comments @ismay. After your explanations, I think the current implementation will work just fine for now. Still think it'd be good to introduce this as a service in the app-runtime too, but that's completely unrelated.

ismay:
> Thanks for replying to my comments @ismay. After your explanations, I think the current implementation will work just fine for now. Still think it'd be good to introduce this as a service in the app-runtime too, but that's completely unrelated.

Definitely, I agree. That would be nice to have. I can see if there's already an issue for it and create it if there isn't.

HendrikThePendric (approved):
Change to `i18n.language` makes sense 🇹🇭

PR #62, opened 2021-01-11
ci(github): add github workflow files

ismay:
The travis CI integrations were disabled for all our repos, this adds the GH actions to replace them.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+    - package-ecosystem: npm
+      directory: '/'
+      schedule:
+          interval: daily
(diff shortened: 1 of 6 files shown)


HendrikThePendric (approved):
Changes are looking good. Just out of interest: did you add/update these files manually, or was this done for you by running `d2-style install project/react`?

ismay:
> Changes are looking good. Just out of interest: did you add/update these files manually, or was this done for you by running `d2-style install project/react`?

Ah yeah that would have been a nice way to do it. No I took the updated files from the usage-analytics-app. I guess using a d2-style install, and then reverting changes to any dotfiles I wanted to keep as-is would have been more straightforward than what I did.

PR #63, opened 2021-01-12
fix(job-form): fix parameter header

ismay:
I accidentally took the header verbatim from the design, but `Monitoring` shouldn't be part of the header, so I simplified it to just `Parameters`, so that it applies everywhere.


HendrikThePendric (approved):

HendrikThePendric:
FYI: approved because I already knew you were in the process of adding `i18n.t()` 😉

PR #64, opened 2021-01-13
fix(infolink): ensure infolink opens in new window

ismay:
https://jira.dhis2.org/browse/DHIS2-10270

Small fix, ensures info link opens in new tab, like it does in the rest of the app.


HendrikThePendric (approved):
Bonus points for `rel="noopener noreferrer"` 🎉

PR #72, opened 2021-01-28
refactor(queries): move queries up

ismay:
https://jira.dhis2.org/browse/DHIS2-10273

Since we don't have caching yet, moving the queries up will allow us to not fetch as often. Otherwise a lot of parts in the app will be fetching data that has already been fetched again, plus the app will feel quite slow and unresponsive.


HendrikThePendric (approved):
I like these changes a lot 👍  so I've approved.

I've only got one suggestion that could be a minor improvement (not critical at all). You are currently using a separate store-context and a bunch of selectors, which take `store` as their first argument. And the code snippet below showcases the pattern we see emerging in the components that rely on the store:
https://github.com/dhis2/scheduler-app/blob/6693fc9f1d34222ad4e97d54dbd53ee6bff0aa28/src/components/FormFields/JobTypeField.js#L20-L21

Now suppose we would not use the concept of "selectors" but instead switch to "custom hooks". Then these hooks could themselves do `const store = useContext(StoreContext)` and it would be easier to use them.

ismay:
> I like these changes a lot 👍 so I've approved.
> 
> I've only got one suggestion that could be a minor improvement (not critical at all). You are currently using a separate store-context and a bunch of selectors, which take `store` as their first argument. And the code snippet below showcases the pattern we see emerging in the components that rely on the store:
> 
> https://github.com/dhis2/scheduler-app/blob/6693fc9f1d34222ad4e97d54dbd53ee6bff0aa28/src/components/FormFields/JobTypeField.js#L20-L21
> 
> Now suppose we would not use the concept of "selectors" but instead switch to "custom hooks". Then these hooks could themselves do `const store = useContext(StoreContext)` and it would be easier to use them.

True, I didn't think of that. Good catch. Should be easy enough to fix, the hooks could just use the selectors internally (or the logic could be moved to the hooks). Created an issue for it here: https://jira.dhis2.org/browse/DHIS2-10451

PR #74, opened 2021-02-08
fix(useeffect): cleanup subscriber

ismay:
Cleaning up the subscriber with useeffect. And added explanatory comments.

### DIFF
--- a/src/components/Forms/JobAddForm.js
+++ b/src/components/Forms/JobAddForm.js
@@ -30,9 +30,14 @@ const JobAddForm = ({
      * Lift pristine state up on changes, wrapped in useEffect because calls to setState
      * outside of the component that owns the setState should not happen synchronously.
      */
-    useEffect(() => {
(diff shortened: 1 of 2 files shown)


HendrikThePendric (approved):
Even though, retrospectively, the comment is clear enough, I still went to the react-final-forms docs to check if a call to `subscribe` does in fact return an unsubscribe function. Perhaps adding a link to the relevant section of the docs would help even more?

https://final-form.org/docs/final-form/types/FormApi#subscribe

ismay:
> Even though, retrospectively, the comment is clear enough, I still went to the react-final-forms docs to check if a call to `subscribe` does in fact return an unsubscribe function. Perhaps adding a link to the relevant section of the docs would help even more?
> 
> https://final-form.org/docs/final-form/types/FormApi#subscribe

Yeah good suggestion, I've added it 👍

PR #75, opened 2021-02-09
fix(editform): replace query with store

ismay:
I missed a query when moving everything to the store. This takes care of the last remaining dataquery.


HendrikThePendric (approved):
Nice!

PR #77, opened 2021-02-11
fix(include-system-jobs): use checkbox for include system jobs switch

ismay:
@cooper-joe mentioned that this switch should be a checkbox, to better match the distinction of switches vs checkboxes in our ecosystem (switches are for turning things on/off, checkboxes for the rest, if I'm quoting that correctly).


src/pages/JobList/JobList.js
```diff
@@ -45,8 +45,8 @@ const JobList = ({
                         value={jobFilter}
                     />
                     <div className={styles.controlRight}>
-                        <Switch
-                            dataTest="job-toggle-switch"
+                        <Checkbox
+                            dataTest="job-toggle-checkbox"
                             checked={showSystemJobs}
                             label={i18n.t('Include system jobs')}
```
cooper-joe, line 51: Would it also be possible to change this label to "Show system jobs", just to make it completely clear that this is about _display_, not whether the job is running/included in any actions.

ismay: Definitely, I'll update it 👍

ismay: Though on second thought, it was `Show system jobs` until recently. The reason I changed it back to `Include system jobs`, is because if you toggle the checkbox, it'll _add_ the system jobs to the list. To me, `Show system jobs` implies it'll show _only_ system jobs, when what it's doing now is including them in the list (though I agree, include does not imply visibility as clearly).

cooper-joe: Ah yeah, good point. In that case, I think we can go for `Include system jobs in list` - that should be completely clear.

ismay: Sounds good 👍

ismay: Updated in 6b59154

HendrikThePendric (approved):
Very Nice

PR #80, opened 2021-02-16
fix(list): move list state to store

ismay:
https://jira.dhis2.org/browse/DHIS2-10453

To solve it I've just moved the state to the store. I'm using hooks as that's what we'll refactor the existing approach to anyway (the selectors).


src/.eslintrc.js
```diff
@@ -61,6 +61,7 @@ module.exports = {
             files: ['*.test.js'],
             rules: {
                 'i18next/no-literal-string': 'off',
+                'react/prop-types': 'off',
```
HendrikThePendric, line 64: Why was this necessary?
And if it was necessary at some place, I would have preferred just some disabling comments there.

HendrikThePendric (approved):

HendrikThePendric (changes requested):
Sorry, I didn't mean to approve before. I'm a bit concerned about that disabled ESLint rule.

src/.eslintrc.js
ismay: This applies just to the test files. Reason I added it is because when you define a wrapper component in a test (like I've done in a couple of tests), eslint will complain if you don't add prop-types for that component.

Now I think the rule is good for client-side code, but I don't think it's useful for tests. So instead of adding disable comments for every test, I've just disabled it globally for all tests. I thought that was cleaner.

HendrikThePendric: Oops I missed the `files` part.

HendrikThePendric (approved):

src/.eslintrc.js
ismay: No worries 👍 Thanks for the review!

PR #107, opened 2021-02-23
fix(jobtype): add new disable inactive users job type 

ismay:
This adds a new jobType that was added on the backend.


mediremi (approved):

Mohammer5 (commented):
LGMT & works

PR #169, opened 2021-04-12

ismay:
Superseded by: https://github.com/dhis2/scheduler-app/pull/171


ismay:
Maybe we could condense this to:

- `push` triggered workflow, runs for everything except branches outside of the repo and dependabot PRs
- `pull_request` triggered workflow. Runs only for branches outside of the repo and dependabot PRs
- release workflow, needs the push workflow and runs only on semantic release branches

If we had job level branch filtering, we could include the release step in the push workflow, but that doesn't seem possible yet: https://github.community/t/run-step-on-if-branch-tag-is/16965

PR #171, opened 2021-04-12

ismay:
Slightly simpler solution than #169 (in my opinion).

Short summary:

- I feel that there's a fair amount of noise in the CI step descriptions (at the bottom of a PR). To make them easier to parse I've:
  -  Removed the dhis2 prefix (this didn't communicate anything extra to me, but maybe I'm missing the reason for this prefix).
  - Removed the app suffix (I can see how it's necessary to distinguish between the different types of workflows when managing multiple workflow types on an org. level, but from the context of this app, I thought it would be clear enough that the workflow running should be for an app).
- Split up the verify task in a `verify trusted` and `verify external` workflow. `verify external` only runs tasks that don't need credentials (and runs for PRs from forks and dependabot PRs). We distinguish between PRs from forks and local PRs by checking a github supplied `fork` boolean. We use a github supplied `actor` to identify dependabot PRs. You can verify that this works correctly by checking the tasks that ran in the commit where I reversed the checks: https://github.com/dhis2/scheduler-app/pull/171/commits/c04c86bb00b30c27caf127b3e39388013d796759. We'll need to check that the dependabot identification works by rebasing the dependabot PRs after merging this.
- Abstracted the publish and release steps to a separate `publish` workflow. Felt like they shouldn't be under verify to me, since we're not verifying but publishing. The `publish` workflow only runs after the `verify trusted` workflow has completed (and jobs are only executed if the `verify trusted` workflow exited successfully)
- [manual skip ci check is no longer necessary](https://github.blog/changelog/2021-02-08-github-actions-skip-pull-request-and-push-workflows-with-skip-ci/), so I removed it
- Unfortunately cypress needs a secret, so this can't run in the `verify external` context
- Note that the publish step won't trigger now. It'll only start triggering once that workflow has been merged to the default branch of this repo.

FYI: I'll squash this to the PR title before merging


ismay:
Closed as I've moved the core fixes from this PR to https://github.com/dhis2/workflows/pull/10

PR #173, opened 2021-04-14
ci(cypress): add a unique cypress id for each e2e run

ismay:
https://jira.dhis2.org/browse/CLI-36

Since I was already working on the workflows, I thought I'd tackle this one while I'm at it. Cypress CI runs can't be restarted with our current config (see https://jira.dhis2.org/browse/CLI-36). This fixes that by passing a unique id for each workflow run.

FYI, for this repo e2e hasn't been enabled yet, so currently the task will fail because of that.

### DIFF
--- a/.github/workflows/dhis2-verify-app.yml
+++ b/.github/workflows/dhis2-verify-app.yml
@@ -105,50 +105,60 @@ jobs:
             - name: Test
               run: yarn test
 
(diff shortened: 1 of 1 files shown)


ismay:
Fixes have been moved to https://github.com/dhis2/workflows/pull/11

PR #206, opened 2021-05-27
chore(deps): update cypress packages

ismay:
Done:

- Add notes to readme about the necessary env vars (I've also removed the outdated info from the readme)
- Live script doesn't exit when closing cypress
- Record flag
- Squash and rebase
- Remove the username and password from the repo secrets (we shouldn't need them)

Follow up:

- https://jira.dhis2.org/browse/CLI-52 (take plugin settings from process.env)
- port e2e workflow updates to workflow repo
- document new settings for the cypress utils in the cypress utils docs (like the network mode)
- update to stable release of cypress utils asap


.github/workflows/dhis2-verify-app.yml
```diff
@@ -101,54 +101,47 @@ jobs:
...
+              id: yarn-cache
+              with:
+                  path: '**/node_modules'
+                  key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+
+            - name: End-to-End tests
+              uses: cypress-io/github-action@v2
+              with:
+                  start: yarn start
+                  command: yarn d2-utils-cypress run --stub --serverMinorVersion 37 --appStart ''
```
varl, line 127:
```suggestion
                  command: yarn d2-utils-cypress run --record --stub --serverMinorVersion 37 --appStart ''
```

HendrikThePendric: Since the `run` command from `cli-utils-cypress` doesn't accept this flag yet, let's not pass it for now.

HendrikThePendric (approved):
Perhaps we should wait until a new major version of `cli-utils-cypress` is released so you don't have to use alpha software in your production app....

But apart from that, I believe the current changes are correct, so I am approving.

As for the follow-up points:
> Remove the username and password from the repo secrets (we shouldn't need them)

I think we need to ask @varl to do this.

> Allow passing a record flag to the underlying cypress command

Could you create a JIRA issue for this?

.github/workflows/dhis2-verify-app.yml
ismay: Yeah I've left it out for now

ismay (commented):
There's a couple of changes that I was thinking about for this PR. I think the current situation with the env vars is a little confusing. I've tried to see what exactly was going on, and created this PR: https://github.com/dhis2/scheduler-app/pull/209.

So I understand that Hendrik's rationale for the changes in alpha.6 are that he'd prefer to keep the base_url settings in a single place. So he doesn't prefer the explicit env var in the `cypress:start` command from the above PR. My concern is that I'm not sure whether the localstorage API for the app-platform is considered an external, stable part of the API, or whether it's considered an implementation detail, and thus subject to change outside of major versions.

If it's a stable API, I'm ok with using the localstorage solution. Though it does seem more opaque to me. To combat that I would suggest we clarify the things we use these env vars for in the docs, as it can get quite confusing.

So my suggestions would be:

- Remove install step (see linked PR)
- Ensure we're using stable APIs (suggestion in linked PR, see above)
- Use uppercase env vars
- See if we can refactor the workflow `command` and `start` commands to use npm scripts so that they can be run locally with the same commands as on CI

.github/workflows/dhis2-verify-app.yml
```diff
@@ -101,54 +101,45 @@ jobs:
...
+            - name: End-to-End tests
+              uses: cypress-io/github-action@v2
+              with:
+                  start: yarn cypress:start
+                  wait-on: 'http://localhost:3000'
+                  wait-on-timeout: 300
+              env:
+                  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+                  CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
+                  # Used by the network shim
```
varl, line 135:
```suggestion
```

.github/workflows/dhis2-verify-app.yml
```diff
@@ -101,54 +101,45 @@ jobs:
...
+
+            - name: Install Cypress binary
+              run: yarn cypress install
+
+            - name: End-to-End tests
+              uses: cypress-io/github-action@v2
+              with:
+                  start: yarn cypress:start
+                  wait-on: 'http://localhost:3000'
+                  wait-on-timeout: 300
```
varl, line 131:
```suggestion
                  wait-on-timeout: 300
                  record: true
```

varl (approved):

.github/workflows/dhis2-verify-app.yml
```diff
@@ -101,54 +101,45 @@ jobs:
...
+
+            - uses: actions/cache@v2
+              id: yarn-cache
+              with:
+                  path: '**/node_modules'
+                  key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+
+            - name: Install Cypress binary
+              run: yarn cypress install
+
```
varl, line 125: Perhaps we don't need to do this either?

```suggestion
```

package.json
```diff
@@ -10,10 +10,11 @@
         "lint:css": "stylelint './src/**/*.css' && prettier './src/**/*.css' --check",
         "format": "d2-style apply && yarn format:css",
         "format:css": "prettier './src/**/*.css' --write",
-        "cypress:open": "d2-utils-cypress open --appStart 'yarn start'",
-        "cypress:run": "d2-utils-cypress run --appStart 'yarn start'",
-        "cypress:capture": "d2-utils-cypress run --capture --appStart 'yarn start' --serverMinorVersion 36",
-        "cypress:stub": "d2-utils-cypress run --stub --appStart 'yarn start' --serverMinorVersion 36"
+        "cypress:start": "BROWSER=none yarn start",
+        "cypress:open": "wait-on 'http://localhost:3000' && cypress open",
+        "cypress:run": "wait-on 'http://localhost:3000' && cypress run",
```
ismay, line 15: @HendrikThePendric, @varl:
  
To me it's a little unclear how we expect users to use these commands locally. Normally `yarn cypress:open` would start the app and then open cypress, whereas now it expects the app to be started separately.

I think aligning the cypress:open and cypress:run scripts with the capture and stub scripts is more consistent (so starting the app for open and run as well).

ismay: Ah I see they're actually used in the stub and capture scripts. So maybe then adding a separate open script for local usage would make sense? Which also starts the app, but runs live, say `cypress:live`?

.github/workflows/dhis2-verify-app.yml
ismay: Yup, removed it.

package.json
ismay: I've added the `cypress:live` script, but this led to a couple of other changes. The scripting we were using for capture and stub didn't work for the `live` script as they wouldn't exit when you'd exit cypress. So you'd have to exit the process in the shell yourself, alongside closing the cypress GUI.

With npm I'd use lifecycle scripts for this (pre, post) but those aren't available in yarn. So I've used [start-server-and-test](https://github.com/bahmutov/start-server-and-test), which is from a cypress maintainer, and recommended in their docs: https://docs.cypress.io/guides/continuous-integration/introduction#Boot-your-server.

Now when running `cypress:open`, when you're done you can close the gui and it'll exit the script and kill the server. I simplified the scripts a little, let me know if you hate it.

varl: I don't care that much, but you could have added `--kill-others` and `--success first` to the concurrently script to do the same thing.

```sh
concurrently 'yarn cy:start' 'yarn cy:run --env networkMode=capture' --kill-others --success first

concurrently 'yarn cy:start' 'yarn cy:run --env networkMode=stub' --kill-others --success first
```

ismay: Ah I see, didn't know that. Yeah I thought I'd go with the one from the cypress docs as the maintainer (bahmutov) mentions correctly handling things like killing the server process etc. Seemed nice and readable and when it worked I didn't look any further.

Plus the maintainer is a ninja.

HendrikThePendric (approved):
Looks great, nice to see this up-and-running.

One suggestion could be to execute this run in parallel mode, which should speed the checks up a lot. But that's entirely up-to-you.

ismay:
> Looks great, nice to see this up-and-running.
> 
> One suggestion could be to execute this run in parallel mode, which should speed the checks up a lot. But that's entirely up-to-you.

Yeah sounds good 👍

varl (approved):

PR #210, opened 2021-06-03
chore: update cypress (viva patch)

varl:
Example for how to align the CI with the new alpha version of `d2-utils-cypress`.

### DIFF
--- a/.github/workflows/dhis2-verify-app.yml
+++ b/.github/workflows/dhis2-verify-app.yml
@@ -132,7 +132,12 @@ jobs:
               env:
                   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
                   CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
+                  # Used by the app platform
+                  DHIS2_BASE_URL: https://debug.dhis2.org/dev
+                  # Used by the network shim
                   CYPRESS_dhis2BaseUrl: https://debug.dhis2.org/dev
+                  CYPRESS_dhis2ApiVersion: 37
+                  CYPRESS_networkMode: stub
 
     publish:
         runs-on: ubuntu-latest
--- a/.github/workflows/dhis2-verify-app.yml
+++ b/.github/workflows/dhis2-verify-app.yml
@@ -132,8 +132,6 @@ jobs:
               env:
                   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
                   CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
-                  # Used by the app platform
-                  DHIS2_BASE_URL: https://debug.dhis2.org/dev
                   # Used by the network shim
                   CYPRESS_dhis2BaseUrl: https://debug.dhis2.org/dev
                   CYPRESS_dhis2ApiVersion: 37
--- a/.github/workflows/dhis2-verify-app.yml
+++ b/.github/workflows/dhis2-verify-app.yml
@@ -127,13 +127,12 @@ jobs:
               uses: cypress-io/github-action@v2
               with:
                   start: yarn cypress:start
-                  command: yarn d2-utils-cypress run --stub --serverMinorVersion 37 --appStart ''
                   wait-on: 'http://localhost:3000'
                   wait-on-timeout: 300
               env:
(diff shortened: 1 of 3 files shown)


ismay (commented):
1. I've added a couple env vars in 88f822c11163b8f31e3f8a54a0ec0811e9d0a8d9:
  a. It didn't seem like the `DHIS2_BASE_URL` for the app-platform was present so I added that
  b. I've also added the api-version and network-mode vars, as it seemed from https://github.com/dhis2/cli-utils-cypress  /pull/218 that those are necessary

2. I guess we could decide to have some of these env vars I've added on CI in the npm scripts. But that would override any cypress.json/cypress.env.json values for our users when running those scripts on their local machines, so I stuck to the above strategy, seemed more flexible.

3. I think we can omit the cypress install step as it's already installed as part of cli-utils-cypress. I did that here as well: https://github.com/dhis2/scheduler-app/pull/209 and it worked fine without.

.github/workflows/dhis2-verify-app.yml
```diff
@@ -127,13 +127,17 @@ jobs:
...
-                  command: yarn d2-utils-cypress run --stub --serverMinorVersion 37 --appStart ''
                   wait-on: 'http://localhost:3000'
                   wait-on-timeout: 300
               env:
-                  cypress_dhis2_base_url: https://debug.dhis2.org/dev
-                  cypress_record_key: ${{ secrets.CYPRESS_RECORD_KEY }}
                   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+                  CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
+                  # Used by the app platform
+                  DHIS2_BASE_URL: https://debug.dhis2.org/dev
```
varl, line 136: I thought that because we call [`setDhis2BaseUrlToLocalStorage`](https://github.com/dhis2/cli-utils-cypress/blob/alpha/packages/cypress-commands/src/helper/dhis2BaseUrl.js#L15) and use the `dhis2BaseUrl` Cypress env variable when we [enable the `autoLogin` plugin](https://github.com/dhis2/cli-utils-cypress/blob/alpha/packages/cypress-commands/src/setups/enableAutoLogin.js#L17-L20) and when we [enable the `networkShim` plugin](https://github.com/dhis2/cli-utils-cypress/blob/alpha/packages/cypress-commands/src/setups/enableNetworkShim/index.js#L33-L35) we don't need to duplicate the `DHIS2_BASE_URL` here?

varl:
> 1. I've added a couple env vars in [e8119d0](https://github.com/dhis2/scheduler-app/commit/e8119d0ab3cf95e0c0feabd38597a6434cccef2e):
>    a. It didn't seem like the `DHIS2_BASE_URL` for the app-platform was present so I added that

See this comment: https://github.com/dhis2/scheduler-app/pull/210#discussion_r644965853

>    b. I've also added the api-version and network-mode vars, as it seemed from https://github.com/dhis2/cli-utils-cypress  /pull/218 that those are necessary

Great, yeah, they are. :+1: 

> 2. I guess we could decide to have some of these env vars I've added on CI in the npm scripts instead. But that would override any cypress.json/cypress.env.json values for our users when running those scripts on their local machines, so I stuck to the above strategy, seemed more flexible.

I prefer adding them to the env in CI too.
 
> 3. I think we can omit the cypress install step in the workflow as it's already installed as part of cli-utils-cypress. I did that here as well: #209 and it worked fine without.

Nice. :+1: 

> 4. When we merge this (or before even) we should remove the username and password secrets from this repo. No longer necessary.

Done.

> 5. Once we've finalized the approach I can document the necessary local and CI env vars for our approach. Seems good to have that somewhere, say in the README. That's always a hassle when cloning a codebase you're not familiar with.

Indeed, that's good to have !

.github/workflows/dhis2-verify-app.yml
varl: Or is it for the `yarn start` command that the Cypress action runs?

varl, line 136:
```suggestion
```

varl: Trying without it in a76a44b and will revert it if it doesn't work.

varl: It works without this env variable, so I think we can leave it out.

varl:
I'll merge this into your update-cypress branch and we can take it from there.

PR #324, opened 2021-11-18
fix: switch to POST for executing jobs

mediremi:
Closes https://jira.dhis2.org/browse/DHIS2-12156

[The HTTP method for the `jobConfigurations/:id/execute` endpoint has switched from `GET` to `POST`](https://github.com/dhis2/dhis2-core/pull/9252).

The `RunJobModal` component has been updated to use `POST`.

### DIFF
--- a/src/components/Modal/RunJobModal.js
+++ b/src/components/Modal/RunJobModal.js
@@ -37,7 +37,12 @@ const RunJobModal = ({ id, hideModal }) => {
             </ModalContent>
             <ModalActions>
                 <ButtonStrip end>
-                    <Button name="hide-modal" secondary onClick={hideModal}>
+                    <Button
+                        name="hide-modal"
+                        secondary
+                        onClick={hideModal}
+                        disabled={loading}
+                    >
                         {i18n.t('Cancel')}
                     </Button>
                     <Button
--- a/src/components/Modal/RunJobModal.js
+++ b/src/components/Modal/RunJobModal.js
@@ -13,10 +13,10 @@ import i18n from '@dhis2/d2-i18n'
 import { hooks } from '../Store'
 
 const RunJobModal = ({ id, hideModal }) => {
-    const [mutation] = useState(() => ({
+    const [mutation] = useState({
         resource: `jobConfigurations/${id}/execute`,
         type: 'create',
-    }))
+    })
     const [runJob, { loading, error }] = useDataMutation(mutation, {
         onComplete: () => {
             hideModal()
--- a/src/components/Modal/RunJobModal.js
+++ b/src/components/Modal/RunJobModal.js
@@ -1,30 +1,39 @@
-import React from 'react'
+import React, { useState } from 'react'
 import { PropTypes } from '@dhis2/prop-types'
-import { useDataEngine } from '@dhis2/app-runtime'
+import { useDataMutation } from '@dhis2/app-runtime'
 import {
     Button,
     Modal,
     ModalContent,
     ModalActions,
     ButtonStrip,
+    NoticeBox,
 } from '@dhis2/ui'
 import i18n from '@dhis2/d2-i18n'
 import { hooks } from '../Store'
 
 const RunJobModal = ({ id, hideModal }) => {
-    const engine = useDataEngine()
-    const query = {
-        jobs: {
-            resource: `jobConfigurations/${id}/execute`,
+    const [mutation] = useState(() => ({
+        resource: `jobConfigurations/${id}/execute`,
+        type: 'create',
+    }))
+    const [runJob, { loading, error }] = useDataMutation(mutation, {
+        onComplete: () => {
+            hideModal()
+            refetchJobs()
         },
-    }
-    const runJob = () => engine.query(query)
+    })
     const refetchJobs = hooks.useRefetchJobs()
 
     return (
         <Modal open small onClose={hideModal}>
             <ModalContent>
-                {i18n.t('Are you sure you want to run this job?')}
+                {error && (
+                    <NoticeBox error title={i18n.t('Error running job')}>
+                        {error.message}
+                    </NoticeBox>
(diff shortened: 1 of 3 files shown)


src/components/Modal/RunJobModal.js
```diff
@@ -1,30 +1,39 @@
...
 
 const RunJobModal = ({ id, hideModal }) => {
-    const engine = useDataEngine()
-    const query = {
-        jobs: {
-            resource: `jobConfigurations/${id}/execute`,
+    const [mutation] = useState(() => ({
+        resource: `jobConfigurations/${id}/execute`,
+        type: 'create',
+    }))
```
ismay, line 19: So I assume that this works because react's useState allows passing a function to setState, which will receive the previous state and has to return the new state. To me the semantics of this are not very clear. Essentially this is doing the same as memoizing the mutation right? Or is there another reason that makes you prefer this approach.

I really think we should settle on one recommended approach for this pattern as an org. btw. This is I think the third pattern I've seen us use and that could get difficult to maintain further down the line.

ismay:
I'll assign this to @Mohammer5 as well as he's there tomorrow and I'm not. I'm happy to finish reviewing this next week if there's no rush, otherwise feel free to tackle this beforehand. The gist looks good to me.

Mohammer5 (changes requested):
I think we should either prevent the modal from being closed while loading to ensure that a potential error can be shown to the user or show the error in an alert (which would be my preference as it'd be consistent with other apps)

ismay:
I think my recommendation would be this pattern: https://github.com/dhis2/scheduler-app/pull/325. I know it's verbose, but to me it seems wrong that we recommend from one lib to not set the query during runtime, and then in other areas implement workarounds to set the query during runtime. The engine was meant as an escape hatch so to me it seems like the stable solution in the meantime. Not ideal, but I think the semantics make more sense than us trying to subvert the recommended pattern from one of our other libs.

Couple of things that would probably be good to tackle later (seems out of scope for this PR):

- Discuss @Mohammer5's suggestion, see whether an alert makes sense for any errors (though I prefer the in context noticebox more)
- Update tests to use the CustomDataProvider instead of mocking
- Remove the store abstraction (not at all related to this PR, but I really want to get rid of it)
- Eventually I do think moving towards testing-lib would be nice. Enzyme is becoming a bit of a chore

Mohammer5 (approved):
If the thing that I found will be deferred, I'll approve this issue

ismay:
I'm still wondering about the question I asked wrt. the choice for useState (see review note above). Also, I think we can remove the mocking from tests that don't click the 'run job' button (like in https://github.com/dhis2/scheduler-app/pull/325). After that's addressed this should be good to go as far as I'm concerned as well.

mediremi:
> I think we should either prevent the modal from being closed while loading to ensure that a potential error can be shown to the user

I'll implement that :+1: 

> I'm still wondering about the question I asked wrt. the choice for useState (see review note above)

Since it seems like there's a strong possiblity that resource parameterisation might be implemented in `app-runtime` this week, I'm happy to keep this PR open until then.

src/components/Modal/RunJobModal.js
mediremi: > So I assume that this works because react's useState allows passing a function to setState, which will receive the previous state and has to return the new state

No this is not related to functional state updates - passing a function to `useState` allows calculating initial value of `mutation` just once. This is similar to `useMemo` except that [`useMemo` is not a semantic guarantee, it's just a performance optimisation](https://reactjs.org/docs/hooks-reference.html#usememo).

mediremi: > I really think we should settle on one recommended approach for this pattern as an org. btw. This is I think the third pattern I've seen us use and that could get difficult to maintain further down the line.

The recommended approach for constants inside a functional component, once we have a DHIS2 hooks lib, will probably be [this `useConst` hook](https://gist.github.com/amcgee/42bb2fa6d5f79e607f00e6dccc733482).

ismay: > No this is not related to functional state updates - passing a function to useState allows calculating initial value of mutation just once.

So why not just pass the desired object to initial state directly? That would also set it only once and skip using a callback. Right?

ismay: In other words, it is my understanding that initial state will only be set once: during state initialization (which makes sense, given the name initial state), and persist as long as the component is mounted. Whether you use a callback or directly pass the desired initial value shouldn't matter. That is my understanding, or am I overlooking something?

mediremi: Yep you're right it shouldn't matter whether it's a callback or passed directly, I think I just did it this way since I started with `useMemo` then remembered that it's not the 'correct' solution. I'll remove the callback :+1:

ismay: Thanks. Yeah I was trying to understand whether I missed something since it confused me a little. Tbh, I think I prefer this solution (or useRef if the setting shouldn't trigger a rerender) over a useConst hook. Especially over maintaining our own react hook lib. But I digress.

> This is similar to useMemo except that useMemo is not a semantic guarantee, it's just a performance optimisation.

Yeah you're right, I forgot about that one. That's a good reason. Thanks for elaborating!

ismay (approved):
> Since it seems like there's a strong possiblity that resource parameterisation might be implemented in app-runtime this week, I'm happy to keep this PR open until then.

I'm good with this being merged with the useState based fix.

Only thing I think I'd do is remove the mocking from tests that don't need it. And setting the useState directly (i.e. with an object). If you want to be completely robust I can imagine that you'd want the dynamic query to respond to changes in the id prop, though practically speaking I don't think that's necessary as the app doesn't do that atm.

Feel free to merge if you're happy with the PR 👍, looks good to me!

PR #325, opened 2021-11-22
fix(run-job): update to POST instead of GET

ismay:
See: https://github.com/dhis2/scheduler-app/pull/324


ismay:
Closed in favour of the approach in #324

PR #336, opened 2022-02-09
fix(job-type-field): filter non-implemented jobTypes

Birkbjo:
The JobType `TRACKER_SEARCH_OPTIMIZATION` is not implemented yet, and thus shows as a blank field in the input selection. The API for implementing this jobType does not seem complete: https://jira.dhis2.org/browse/DHIS2-12048 . This prevents future new non-implemented jobTypes from showing as blank fields in the selector. 

Also added sorting of the Job-types. A bit debatable if this should be added here, but I think it's a good quality of life improvement.


ismay (approved):
Looks good!

Mohammer5 (approved):

PR #348, opened 2022-02-25
feat(data-integrity): add checks-selector

Birkbjo:
https://jira.dhis2.org/browse/DHIS2-12286

Don't be too intimidated of all the file changes, most are updated fixtures for tests.

Should probably create a generic field component for 'java.lang.Set', however we still need to override the `TransferOption` due to the label not supporting a react-component. I plan to update this in UI though. 

Note that the https://debug.dhis2.org/dev/api/dataIntegrity.json. endpoint contains properties like `description`, `recommendation` and `introduction`. 

These are very inconsistent, and most jobs do not have all these properties - and `description` seem to just be a human readable `name`. After a brief discussion with Joe, we decided a simple MVP is good enough for now. So we opted to just show the `name` and `severity`. 

This endpoint is "semi-dynamic", in the sense that it's described by a `yaml`-file, but still compiled at build-time, it also does not have support for translations. I've included translations for all the current jobs, but also have a simple fallback for any other jobs with a simple "snake"-case converter. This should all ideally be done on the backend, but we do what we have to work with what've got.


src/components/FormFields/DataIntegrityChecksField.js
```diff
@@ -0,0 +1,183 @@
...
+                    label={label}
+                    validate={VALIDATOR}
+                />
+            )}
+        </FieldGroup>
+    )
+}
+
+// Mostly taken from https://github.com/dhis2/ui/blob/master/components/transfer/src/transfer-option.js
+// TODO: Change PropType of label in ui to accept "node" so this won't be necessary
```
Mohammer5, line 84: It's just a prop types issue, right? It wouldn't be shown in production, so we could just ignore it and add a follow up Jira issue to fix the prop type in UI

mediremi (approved):
LGTM :heavy_check_mark:

src/components/FormFields/DataIntegrityChecksField.js
```diff
@@ -0,0 +1,158 @@
+import React, { useCallback, useRef, useState } from 'react'
```
mediremi, line 1:
```suggestion
import React, { useCallback, useState } from 'react'
```

src/components/FormFields/ParameterFields.js
```diff
@@ -22,20 +39,27 @@ const ParameterFields = ({ jobType }) => {
     }
 
     // Map all parameters to the appropriate field types
-    const parameterComponents = parameters.map(({ fieldName, name, klass }) => {
+    const parameterComponents = parameters.map(({ fieldName, name, klass, ...rest }) => {
         const defaultProps = {
             label: fieldName,
             name: `${FIELD_NAME}.${name}`,
         }
+        const parameterProps = {
```
Birkbjo, line 47: Abit unsure about this, but wanted to keep the raw `properties` from the `job-configuration`, and not conflict with label and `name` above.

cypress.json
```diff
@@ -3,5 +3,10 @@
     "projectId": "sc56ms",
     "testFiles": "**/*.feature",
     "experimentalInteractiveRunEvents": true,
-    "video": false
+    "video": false,
+    "env": {
+        "dhis2DataTestPrefix": "dhis2-scheduler",
+        "networkMode": "live",
+        "dhis2ApiVersion": "37"
```
ismay, line 10: Should this not also be 38?

src/services/server-translations/dataIntegrityChecks.js
```diff
@@ -0,0 +1,123 @@
...
+    SEVERE: i18n.t('Severe'),
+}
+
+const reportTypeMap = {
+    REPORT: i18n.t('Report'),
+    SUMMARY: i18n.t('Summary'),
+    DETAILS: i18n.t('Details'),
+}
+
+const snakeCaseToHumanReadable = string => {
```
ismay, line 110: Is this so that untranslated string will look better in english? Because I can imagine that this would also make it harder to catch untranslated strings, as we usually look at the UI in english (or I do at least).

src/components/FormFields/ParameterFields.js
```diff
@@ -6,13 +6,33 @@ import { hooks } from '../Store'
...
+    ) {
+        return DataIntegrityReportTypeField
+    } else if (parameterName === 'skipTableTypes') {
+        return SkipTableTypesField
+    }
+    return null
+}
+
 // Renders all parameters for a given jobtype
 const ParameterFields = ({ jobType }) => {
```
ismay, line 37: I think these changes go somewhat against the grain of the design of the app (just like the skipTableTypes parameter). The app is meant to respond to the `klass` types (however ugly that is) and render the appropriate fields.

I see that the backend returns a `java.util.Set` type for the `Checks` parameter. We used to have a field for that type in the app that was removed because it was unused. Would it be sufficient to assume that the Set type indicates the transfer component? Or is this control so specific that it can't be solved in a generic manner? Because that would align with the existing design and hopefully then also generalize to other parameters that might need a transfer component.

Also, the `org.hisp.dhis.scheduling.parameters.DataIntegrityJobParameters$DataIntegrityReportType` klass for the `Type` field seems very inconsistent with the other klass fields. This seems to just render a single select. For other parameters we use a `java.util.List` for that. Is there a reason that the backend doesn't just return a `java.util.List` there?

ismay: I'm not saying that I like the current design of dynamically responding to the klass fields btw. I think we should move to fixed forms like the usage-analytics-app. But until we do that I think it can get confusing if we start mixing approaches.

src/components/FormFields/DataIntegrityChecksField.js
```diff
@@ -0,0 +1,158 @@
...
+
+const ChecksTransfer = ({ input, meta, options = [], hidden }) => {
+    const { onChange } = input
+
+    const handleChange = useCallback(
+        ({ selected }) => {
+            onChange(selected)
+        },
+        [onChange]
+    )
```
ismay, line 116: I'm curious, why not use an inline fn in the transfer's onchange directly?

src/components/FormFields/DataIntegrityChecksField.js
```diff
@@ -0,0 +1,158 @@
...
+                height={'450px'}
+                selectedEmptyComponent={<SelectedEmptyComponent />}
+                className={styles.transfer}
+            />
+            {isErr && <Help error={isErr}>{meta.error}</Help>}
+        </>
+    )
+}
+
+ChecksTransfer.propTypes = InputFieldFF.propTypes
```
ismay, line 143: This will prevent eslint from linting the proptypes. It'll only lint them if they're defined directly, last I checked.

src/components/FormFields/DataIntegrityChecksField.js
```diff
@@ -0,0 +1,158 @@
...
+            onChange([])
+        }
+        setRunSelected(checked)
+    }
+
+    return (
+        <FieldGroup label={i18n.t('Checks to run')}>
+            <Radio
+                name={'checksToRun'}
+                value={String(false)}
```
ismay, line 60: Why not just write out "false"? Maybe I'm missing something?

src/components/FormFields/DataIntegrityChecksField.js
```diff
@@ -0,0 +1,158 @@
...
+            label: getCheckName(option.name),
+            severity: severityMap[option.severity],
+        }))
+        .sort((a, b) => a.label.localeCompare(b.label))
+
+    const toggle = ({ value }) => {
+        const checked = value === 'true'
+
+        if (!checked) {
+            // clear checks when "Run all" is selected
```
ismay, line 50: We have destroyOnUnregister enabled, see [here](https://github.com/dhis2/scheduler-app/blob/6915b1858a13bcfaf4486edd76cd1789a3871ff6/src/components/Forms/JobEditFormContainer.js#L50), which should take care of removing the value if the field is unregistered. But I see you linked an open issue about conditional rendering causing issues. We're rendering the parameter fields conditionally, could what you encountered be solved in a similar manner?

ismay:
Ok, I see some of my questions were already answered in the PR description. Sorry about that. I'll go through them and close what you already explained.

src/services/server-translations/dataIntegrityChecks.js
ismay: >This endpoint is "semi-dynamic", in the sense that it's described by a yaml-file, but still compiled at build-time, it also does not have support for translations. I've included translations for all the current jobs, but also have a simple fallback for any other jobs with a simple "snake"-case converter. This should all ideally be done on the backend, but we do what we have to work with what've got.

I think this explains it. I'll close this as it seems like we should discuss this with backend.

src/components/FormFields/ParameterFields.js
ismay: > Should probably create a generic field component for 'java.lang.Set', however we still need to override the TransferOption due to the label not supporting a react-component. I plan to update this in UI though.
>
> Note that the https://debug.dhis2.org/dev/api/dataIntegrity.json. endpoint contains properties like description, recommendation and introduction.
>
> These are very inconsistent, and most jobs do not have all these properties - and description seem to just be a human readable name. After a brief discussion with Joe, we decided a simple MVP is good enough for now. So we opted to just show the name and severity.

I think this explains most of it. So basically the transfer component needs specifics that can't really be expressed with just the `Set` klass type. Seems that also here my questions should be directed to backend.

cypress.json
Birkbjo: Yeah, I would think so as well... Hendrik added this, but I guess the param from the command takes precedence, but will fix.

src/components/FormFields/DataIntegrityChecksField.js
Birkbjo: True, I tried a few different approaches here, since at first I wasn't aware that value had to be string. Can of course just be "false".

src/components/FormFields/DataIntegrityChecksField.js
Birkbjo: Well, I got errors that crashed the app, so I added the `hidden`-prop to fix this. It wasn't just because of conditional rendering, but conditional rendering in combination with updating the state of the component.

src/components/FormFields/ParameterFields.js
Birkbjo: I know that you resolved this, but I will just add my thoughts here :)

> I think this explains most of it. So basically the transfer component needs specifics that can't really be expressed with just the Set klass type. Seems that also here my questions should be directed to backend.

Exactly, I've also looked into a generic "transfer"-component in a separate branch, that we can use for an upcoming feature. I also tried to look into adding pagination to this, but that is very hacky at the moment due to the `runtime` not supporting dynamic queries. 

> I think we should move to fixed forms like the usage-analytics-app.

This is something we can discuss, but there's definitely some advantages to have some dynamic rendering of the forms. This is something that will come up in regards to the Maintenance refactor, so we will probably discuss it a lot then. 

In regards to your points about this being "against the design of the app"; In any situation where we implement something dynamic from the server there has to be ways to override them. Of course, in an ideal world we would get all we need from the server and just render it. However, we cannot expect to ever get enough information to always be able to rely on that, and there will always be anomalies - eg; how this transfer component uses custom rendering of the options. And I don't think this is a bad thing. The Maintenance app gets a lot of flak, but some of the great parts is how easy it is to add new fields to due the dynamic resolution of fields. 
Of course if we move to more fixed fields, all this doesn't really matter. In the meantime there's not really an alternative here. It doesn't really have to be confusing either, as we could integrate the `custom`-fields solution better, but I didn't want to change too much. 

> org.hisp.dhis.scheduling.parameters.DataIntegrityJobParameters$DataIntegrityReportType

I was a bit confused by this as well, but I felt that I had already bothered the backend dev enough already, and it was after soft freeze. I think it has to do with the `constants`-array that is included in the payload. We are also using the `name` and `jobType`-props to resolve the field regardless.

src/components/FormFields/DataIntegrityChecksField.js
Birkbjo: Think I did this because I was debugging some re-renders and weird behaviour for the transfer-component, and just left it as is. I mean, it's probably not necessary, but I'm also curious why you would prefer inline, as this just means potentially less re-renders?

src/components/FormFields/DataIntegrityChecksField.js
Birkbjo: The error I encountered was ```Cannot update a component (`DataIntegrityChecksField`) while rendering a different component (`ForwardRef(Field)`). To locate the bad setState() call inside `ForwardRef(Field)`,```

src/components/FormFields/DataIntegrityChecksField.js
ismay: These articles have always been what I personally refer to with regards to inline functions react components: 

- https://medium.com/@ryanflorence/react-inline-functions-and-performance-bdff784f5578 
- https://kentcdodds.com/blog/usememo-and-usecallback.

It doesn't always make things better so I just refrain from using them for performance reasons until there's a clear need for them.

src/components/FormFields/DataIntegrityChecksField.js
ismay: I see. It seems like you're initializing the state for `runSelected` based on the previous value of the field. Which will be there if the user toggles back and forth between "run all" and "run only". It'll be lost as soon as the user switches to another job type. If you lose that and initialize `runSelected` to `false` there's no longer a need to clear the onChange (provided you conditionally render the transfer field in the field group). That means you shouldn't run into the linked bug either (at least I didn't during testing, but let me know if I missed it).

I'd go with that. We don't save state for parameter settings between jobs anyway, and if not saving it for this section doesn't require workarounds and simplifies things I would personally go in that direction.

ismay: Haha, oh yeah of course you're right. Minor oversight on my part 😅. Wait, let me check one thing that might still satisfy that requirement.

ismay: Ok. So it seems that using useField prevents the field from unregistering. That's why you need to clear manually even if you're no longer rendering the field.

Using useFormState allows access to the form values without keeping the field registered:

```
    const { values } = useFormState({ subscription: { values: true } })
    const value = values?.jobParameters?.checks

    // ...
    
    const toggle = ({ value }) => {
        const checked = value === 'true'
        setRunSelected(checked)
    }

    // ...

            {runSelected && <Field
                name={name}
                component={ChecksTransfer}
                options={translatedOptions}
                label={label}
                validate={VALIDATOR}
            />}
```

Accessing the field from FF state might get a little ugly because the name uses final form's notation for nested fields. But something like this would keep the desired behaviour and allow you to do conditional rendering without encountering the bug (at least I didn't with this approach).

Birkbjo: Thanks for looking into this! Actually makes sense for `useField` to prevent unregistering of the field. However, I don't really think this is much better than the current solution? This is a bug in `react-final-form` anyway, and it's referenced in the code - so once it's fixed we could move the "hidden"-prop and conditionally render instead. I don't really see the benefit here - you would even have more re-renders due to subscribing to all values (I know that the current code also re-renders for each change, but that is easily preventable with `react.memo()` if we want that). 

> I do wonder if our usage of initialValues in the app is a little flawed. They'll be lost as soon as final form unregisters a field and clears the value, like when a user switches jobTypes for an existing job.

Yeah, haven't looked too much into that, but seems like we would need to change the `destroyOnUnmount` or something if we would like to keep the initialValues when unmounting fields?

ismay: > Yeah, haven't looked too much into that, but seems like we would need to change the destroyOnUnmount or something if we would like to keep the initialValues when unmounting fields?

Yeah, yeah the reason that that's there is so that if you switch a jobType, the existing parameters don't persist. Otherwise those parameters will be sent to the backend if the user submits. I suppose the backend would probably drop unrecognized job parameters, but it seemed more correct to me at the time.

That's also why I was thinking of a way to automatically remove the transfer values if it unregisters. My assumption was that selecting run all would mean you don't need the checks job parameter at all. But of course, run all is identical to selecting all checks (I assume?). Which is obvious of course, but I'd not even stopped to consider that.

So, first off, I'll approve the PR since I don't want to block you. Since I rewrote a lot of this app I tend to get a little product-managery about it, but that's not actually my position.

That being said. I can't help myself thinking that the design could be simplified. Instead of a "run all" radio button I'd just render the transfer at all times and forego the radios. After all, the transfer already has a convenient "select all" control. That would simplify our code _and_ the UI if you ask me. Do you know if there's a technical reason why we didn't go with that approach?

ismay (approved):

src/components/FormFields/DataIntegrityChecksField.js
Birkbjo: > So, first off, I'll approve the PR since I don't want to block you. Since I rewrote a lot of this app I tend to get a little product-managery about it, but that's not actually my position.

I absolutely understand this, no worries! There's a lot of decision that's taken here, and I don't know the full background. I've just tried to implement this feature without too drastic changes. 

> That being said. I can't help myself thinking that the design could be simplified. Instead of a "run all" radio button I'd just render the transfer at all times and forego the radios. After all, the transfer already has a convenient "select all" control. That would simplify our code and the UI if you ask me. Do you know if there's a technical reason why we didn't go with that approach?

I do agree with this point, and this was my first implementation, with a "If no checks are selected, all checks will be run"-message in the "SelectedEmptyComponent". But I talked to Joe, and he proposed the implemented solution. The reason being that it might be a bit confusing that actually selecting nothing selects all, and the reason below:
Manually having to select all would also probably not be the best, since if new checks are added in the backend, you would have to manually select them again. In the current solution, all checks will run regardless of user interaction after the fact.

ismay: I see. Since we already talked a bit on slack, I'll just add here what I said there: seems tricky to really simplify the UI then on our end. The requirements kind of dictate the current solution. Would be nice if eventually we could get rid of the two types of "all checks", because now effectively we have "all current and future checks" and "all explicitly selected checks". To me that wasn't really clear from the UI. Anyway, something for the future I guess. Thanks for explaining!

PR #377, opened 2022-05-11
ci: replace semantic commit checks with new workflow

ismay:


HendrikThePendric:
This PR only contains formatting changes, so will merge without QA approval

PR #417, opened 2022-09-15
ci(netlify): add production deploy workflow

ghost:
This adds a workflow that will deploy master to netlify on push, as we discussed yesterday. One other option would be to create a branch deploy workflow, with a configurable set of branches. But in going through the netlify action repo it seems that that is not as robust as this method, so I decided to trial this first. If necessary we can always add a 'netlify deploy branch' workflow.

This should deploy to: dhis2-scheduler.netlify.app
An optional brach deploy workflow would deploy to: [branch]--dhis2-scheduler.netlify.app

I've renamed the workflows so the names are cohesive. `Preview production` didn't really make as much sense to me as the name for the new workflow, so I went with `netlify deploy production` and `netlify deploy pr`

### DIFF
--- a/.github/workflows/dhis2-netlify-deploy-pr.yml
+++ b/.github/workflows/dhis2-netlify-deploy-pr.yml
@@ -34,13 +34,16 @@ jobs:
 
             - name: Deploy
               id: netlify-deploy
-              uses: nwtgck/actions-netlify@v1.1
+              uses: nwtgck/actions-netlify@v1.2.2
               timeout-minutes: 1
               with:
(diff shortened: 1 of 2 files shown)


amcgee:
Nice @ismay - shouldn't it be possible to use the same workflow for both branch and production/master deploys by evaluating conditions for those action parameters?  I know that `dev` branch deploys are going to be often requested with apps on continuous delivery workflow

HendrikThePendric (approved):
LGTM. I don't mind have some duplication here since these are just templates that people are meant to edit, so readability is key. The rename is also fine with me.

amcgee (approved):

PR #419, opened 2022-10-04
fix(system-jobs): always allow running jobs manually

ismay:
https://dhis2.atlassian.net/browse/TECH-1300

Before this fix we used to check whether a job was configurable and show a manual run option only if it was. Currently we don't want to block manual runs for system jobs anymore, so I've removed the condition. Manual run will be available for all types of jobs.

The dialog for system jobs before this fix:

image

And after this fix:

image

Triggering a manual run of a system job will send this kind of POST:

image


Mohammer5 (approved):
LGTM

---

I can see that the `configurable` condition still applies to whether a job can be deleted. Is that still correct, system jobs can't be deleted?

PR #426, opened 2022-11-07
fix(validate cron): allow ranges in fraction numerators

Mohammer5:
Closes DHIS2-14016 ([link](https://dhis2.atlassian.net/browse/DHIS2-14016))

### DIFF
--- a/src/services/validators/validate-cron.js
+++ b/src/services/validators/validate-cron.js
@@ -90,6 +90,11 @@ const isValidWeekdayField = (field) =>
  * Format: <second> <minute> <hour> <day-of-month> <month> <day-of-week>
  * Source code: https://git.io/vpoqG
  * Documentation: https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/scheduling/support/CronSequenceGenerator.html
+ *
+ * @TODO: evaluate if the following is possible
+ * There's a npm library called `cron-validate`. It'd need to be configured to
+ * accept the quartz job scheduler syntax, it can't be used as a drop-in
+ * replacement!
  */
 
 const validateCron = (exp) => {
--- a/src/services/validators/validate-cron.js
+++ b/src/services/validators/validate-cron.js
@@ -94,7 +94,7 @@ const isValidWeekdayField = (field) =>
  * @TODO: evaluate if the following is possible
  * There's a npm library called `cron-validate`. It'd need to be configured to
  * accept the quartz job scheduler syntax, it can't be used as a drop-in
- * replacement!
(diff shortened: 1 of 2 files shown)


src/services/validators/validate-cron.test.js
```diff
@@ -35,4 +35,8 @@ describe('validateCron', () => {
     it('should allow for number ranges', () => {
         expect(validateCron('0 0 1-5 * * *')).toEqual(true)
     })
+
+    it('should allow ranges in fraction numerators', () => {
+        expect(validateCron('0 0 10-22/2 ? * *')).toBe(true)
```
ismay, line 40: toBe seems better here than toEqual since we're comparing primitives. To prevent confusion I'd replace all toEqual's here with toBe.

src/services/validators/validate-cron.js
```diff
@@ -88,6 +90,11 @@ const isValidWeekdayField = (field) =>
  * Format: <second> <minute> <hour> <day-of-month> <month> <day-of-week>
  * Source code: https://git.io/vpoqG
  * Documentation: https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/scheduling/support/CronSequenceGenerator.html
+ *
+ * @TODO: evaluate if the following is possible
+ * There's a npm library called `cron-validate`. It'd need to be configured to
+ * accept the quartz job scheduler syntax, it can't be used as a drop-in
+ * replacement!
```
ismay, line 97: I'd drop this in a jira issue I think. Otherwise we'll probably forget about it.

src/services/validators/validate-cron.js
```diff
@@ -40,7 +40,9 @@ const isValidFraction = (fraction, x, y) => {
 
     /* istanbul ignore next */
     return (
-        (isWildcard(components[0]) || isValidNumber(components[0], x, y)) &&
+        (isWildcard(components[0]) ||
+            isValidNumber(components[0], x, y) ||
+            isValidNumberRange(components[0], x, y)) &&
         isValidNumber(components[1], x, y)
```
ismay, line 46: This is a very hard to read section btw. Not your doing since this whole file has logic like that, but it's definitely in need of us cleaning it up.

src/services/validators/validate-cron.js
Mohammer5: I created a jira issue and added the link to the @TODO comment

src/services/validators/validate-cron.js
Mohammer5: Agreed. Out of scope of this PR

ismay (approved):

HendrikThePendric (approved):

PR #452, opened 2023-01-24
fix: update timestamps [DHIS2-14516]

tomzemp:
See [DHIS2-14516](https://dhis2.atlassian.net/browse/DHIS2-14516)

**before**
the time was displayed as a relative time, but we assumed the timestamps from the server were in UTC. So for example, if my server timezone is in Europe/Oslo and it is currently 19:41 and my next job is scheduled at 45 minutes past the hour (19:45), this would display as `in 1 hour` because our code (+ moment) adjusted the timestamp for the assumed 1 hour time difference between CET+1 and UTC timezones

image

**after**
time stamps are displayed as absolute time (`YYYY-MM-DD HH:MM:SS (server timezone)`) with the relative time in a tooltip (this has been okayed by Joe and Jan B., see ticket)
image

[DHIS2-14516]: https://dhis2.atlassian.net/browse/DHIS2-14516?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


geethaalwan (approved):
Tested successfully on 2.40 version

HendrikThePendric (approved):

package.json
```diff
@@ -5,8 +5,8 @@
         "build": "d2-app-scripts build",
         "start": "d2-app-scripts start",
         "start:nobrowser": "BROWSER=none d2-app-scripts start",
-        "test": "d2-app-scripts test --coverage",
-        "test:watch": "d2-app-scripts test --watch",
+        "test": "TZ=Etc/UTC d2-app-scripts test --coverage",
+        "test:watch": "TZ=Etc/UTC d2-app-scripts test --watch",
```
ismay, line 9: I would just mock this in the test instead of globally. Otherwise it's not clear why this is being mocked.

ismay: See how we do it here for example: https://github.com/dhis2/ui/blob/master/components/select/src/multi-select/__tests__/selection-list.test.js#L10-L29

ismay (commented):
Maybe I missed this discussion, but why did we change from relative to absolute time? Relative time seems simpler to parse for a user to me. We could have the actual time in a tooltip as the more detailed option.

edit: ok, I see I did miss some discussion, and that Joe and Jan B. ok'ed this. Tbh., I would find the inverse more user friendly, but if this is not up for discussion at this stage I'll defer to the existing consensus. It just doesn't seem very user friendly to me as the information that seems easier to me to understand (the relative time) is presented as the secondary option.

src/components/JobTable/NextRun.js
```diff
@@ -1,22 +1,34 @@
...
     /**
      * The recommendation is to run dhis2 on a server set to UTC time.
      * In that case this timestamp is also UTC. If those recommendations
      * weren't followed the time could be off, but there's nothing
      * we can do to detect that.
      */
-    const nextRun = moment.utc(nextExecutionTime)
-    const nextRunIsInPast = nextRun.isSameOrBefore(now, 'minute')
+    const nextRun = fromServerDate(nextExecutionTime)
+    const nextRunIsInPast = nextRun.getTime() <= now
```
ismay, line 31: Is this comment still accurate now that we're converting between timezones? It should correct for a server set in a different TZ right?

src/components/JobTable/NextRun.test.js
```diff
@@ -1,24 +1,50 @@
...
 // Z is the zone designator for the zero UTC offset
 const now = new Date('2010-10-10T10:10:10.000Z').valueOf()
 
-// The server does not specify a timezone, but is assumed to provide UTC
 const past = '2009-10-10T10:10:10.000'
 const future = '2011-10-10T10:10:10.000'
+const timeZone = Intl.DateTimeFormat().resolvedOptions().timeZone
+
+// in this test, we are only passing timestamps to the hook (and there is no server/client time diff)
+// so mock getServerZonedISOString by returning the original timestamp
```
ismay, line 14: Have we adequately covered all expected scenarios in this test? Should we not also test for the expected behaviour in case there is a difference between the client and server TZ?

tomzemp: I thought about that originally but then decided against adding tests for different time zones. I guess I thought it would be a bit counter to the goals of the hook, i.e. we want to provide a hook that abstracts away the logic (and responsibility) for adjusting for client/server time zones. 

Hence, I figured that the tests in app-runtime should be sufficient for testing those differences and then a consuming app should only be responsible for testing for a given time zone / testing the "base" functionality without worrying about the client/server time zone difference?

ismay: Yeah I agree, that it's not really necessary to test external library functionality. From the test coverage report it also seems like we've covered everything in NextRun so we should be good 👍

src/components/JobTable/NextRun.js
ismay: What I'm saying is, this could be removed, right?

> If those recommendations
>     * weren't followed the time could be off, but there's nothing
>     * we can do to detect that.

Since we're now detecting TZ differences and correcting for them. Or not?

tomzemp: Yes; going to remove 👍 (when I get around to updating the PR)

package.json
tomzemp: The time zone is being mocked as part of the mock of the `useTimeZoneConversion` hook, so we don't actually need to set it generally for the test (I think this is left over from some other approach I did not use). I've just removed this.

src/components/JobTable/NextRun.js
tomzemp: Updated the comment to say `Adjust for client/sever time zone difference.` (rather than say we cannot detect it)

PR #463, opened 2023-02-22
chore(cypress): set capture to update against debug/dev

ismay:
This branch reproduces an error that occurs when you try to update the cypress fixtures. Running `cypress:capture` will fail with the following error:

image

For #461 I'm updating the fixtures to the latest API version, but this is blocking that update.

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -6,7 +6,7 @@
     "video": false,
     "env": {
         "dhis2DataTestPrefix": "dhis2-scheduler",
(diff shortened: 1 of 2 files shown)


ismay:
Added [d21dc99](https://github.com/dhis2/scheduler-app/pull/463/commits/d21dc99895c45c0baf43d023572a957472472ef3) just for debugging. The error still persists after those changes as well.

PR #464, opened 2023-02-22
test(cypress): use cypress session for authentication

ismay:
We could use this to login during e2e tests and allow for easy switching between users. Some other potential changes we could make:

- Test against the netlify deploy with the GH action deployment_status event
- Test against a dockerized backend (see: https://dhis2.slack.com/archives/C0BP0RABF/p1677502943481919, https://github.com/dhis2/scratch-actions-sandbox/pull/2, https://github.com/dhis2/dhis2-core/blob/master/.github/workflows/run-api-tests.yml#L58)

It does seem like this needs a more recent cypress version than what we're currently supplying through our cypress libs though.


ismay:
Superseded by https://github.com/dhis2/scheduler-app/pull/484

PR #468, opened 2023-02-28
chore: update browserslist, translations and tests

ismay:
This PR bundles some of the updates I've cherry picked from what I'm working on. Two minor things (the browserslist update and the translations). The main thing is the updates to the e2e tests for 2.40.

PS: If merging, please use a regular or rebase merge (squash wouldn't make sense for these commits).

Closes #463


HendrikThePendric (approved):

tomzemp (approved):
lgtm 👍

(though noted and discussed with @ismay that some of the tests appear now to be flaky presumably because of some cypress issue)

ismay:
> (though noted and discussed with @ismay that some of the tests appear now to be flaky presumably because of some cypress issue)

Yeah I was encountering several issues when updating and fixing the e2e tests. Updating the fixtures against debug.dhis2.org failed with a rather cryptic error, but updating against a local instance did work. Also, some of the tests seem rather flaky, as Tom mentions.

My suspicion is that we're hitting https://github.com/cypress-io/cypress/issues/7306, but I'm not sure. Since the cypress version needs to be updated upstream (in our cypress libs) and can't be updated independently it's hard to verify whether this is what we're encountering.

Maybe our tests aren't completely independent (i.e. some leftover state influences the next test), but that seems unlikely to me. All calls that influence the backend should be intercepted (but it's possible I've missed something of course). The tests also do not exhibit this flakiness when testing against 2.38. Plus they do not always fail, even when running them in the exact same order against 2.40. If there's codependence between tests I'd expect them to fail in the same manner if run in the same order from a clean slate, and that does not seem to be the case.

I can imagine it might also be some kind of caching issue (say react-query or a service worker issue), but that's just speculation.

At any rate this re-enables the e2e tests. From here we can debug and improve things further. My recommendation would be that we update cypress to the latest version and check whether that's solved this issue.

PR #484, opened 2023-03-13
test(cypress): update to cypress 12

ismay:
CI for this app is rather flaky. It was blocking the work I'm doing atm on the job queues. This PR fixes a couple things that during my tests have fixed the flakiness.

- Update cypress to v12. For this I've resorted to testing directly against a dockerized backend. It adds about 1m50s overhead to the e2e test run, but removes a lot of complexity. Cypress 12 also allows us to use the new [cy.session](https://docs.cypress.io/api/commands/session) and a simple login helper. We can reuse the session without running too much risk of tests contaminating eachother. Version 12 of cypress also fixes [an issue with tests failing due to rerenders during an assertion](https://github.com/cypress-io/cypress/issues/7306).
- The cypress-cucumber-preprocessor we were using is [no longer maintained](https://github.com/badeball/cypress-cucumber-preprocessor/issues/689). As part of the migration to cypress 12 I've updated to the fork that is the successor: [@badeball/cypress-cucumber-preprocessor](https://github.com/badeball/cypress-cucumber-preprocessor). The configuration it requires is a little different, but nothing noteworthy.

Env vars needed are: 

- `CYPRESS_LOGIN_NAME`
- `CYPRESS_LOGIN_PASSWORD`
- `CYPRESS_LOGIN_SERVER`

See `cypress.env.json.example`. CI is already configured with the necessary env vars. 

This approach is a little novel, but it works well and removes a lot of maintenance overhead for our cypress setup. Plus we're testing against an actual backend so there's also some added realism. See https://dhis2.slack.com/archives/C1XTS67ME/p1678701948371199 for more context.

## Running time

Mocked backend:

- 2m53s with a mocked backend

Pre-seeded image (see here: https://github.com/dhis2/scheduler-app/pull/489):

- 5m25s without any caching (about 2m30s more than the mocked backend)

Pre-seeded image with 4 extra runners:

- 4m40s without any caching (about 1m50s more than the mocked backend)

## Running time other methods

The test code for these runs isn't completely identical to the tests above, login for this suite is slightly slower (about 30s) due to logging in via the ui, so not a completely fair comparison. But these methods were slower overall anyway, so not that important.

Pre-seeded image, layer caching:

- 8m13s with layer caching, empty cache
- 6m45s with layer caching, primed cache

Dockerized backend:

- 7m5s against a dockerized backend, without a pre-seeded db image


package.json
```diff
@@ -32,15 +31,13 @@
         "styled-jsx": "^4.0.1"
     },
     "devDependencies": {
+        "@badeball/cypress-cucumber-preprocessor": "^15.1.4",
+        "@cypress/webpack-preprocessor": "^5.17.0",
```
ismay, line 35: Shouldn't this already be provided by cli-utils-cypress?

PR #495, opened 2023-04-25
refactor: remove store

ismay:
Removing the store and moving to individual data fetching hooks to prepare for the new endpoints.

Preparing for:

- https://dhis2.atlassian.net/browse/DHIS2-14749
- https://dhis2.stoplight.io/docs/dhis2/ehaxt2lzojavb-dhis-2-api-scheduler-only

Follow up

- Remove unnecessary mocking from tests
- Fix deprecated query types (disabled pagination, no fields specified)
- Improve parameter field handling (remove exceptions, could be solved by making the app more static)
- Continuous analytics table does not seem to retain delay
- Reenable skipped tests (ui update might fix it)
- Make loading failures for form fields clearer (use noticebox)
- Split useParameterOptions into separate hooks (could also be solved by making the app more static)


src/components/JobTable/Actions.js
```diff
@@ -17,8 +17,10 @@ const Actions = ({ id, configurable }) => (
                 ) : (
                     <ViewJobAction id={id} />
                 )}
-                {configurable && <RunJobAction id={id} />}
-                {configurable && <DeleteJobAction id={id} />}
+                {configurable && (
+                    <RunJobAction enabled={enabled} id={id} refetch={refetch} />
```
Mohammer5, line 21: Very minor: I think it'd be clearer here if the prop wasn't called `refetch` but `onComplete`.
Then this line would be more understandable: `<RunJobAction ... onComplete={refetch} />`

src/components/JobTable/Actions.js
```diff
@@ -17,8 +17,10 @@ const Actions = ({ id, configurable }) => (
...
+                )}
+                {configurable && <DeleteJobAction id={id} refetch={refetch} />}
```
Mohammer5, line 23: Same as above

src/hooks/job-schedules/index.js
```diff
@@ -0,0 +1,3 @@
+import useJobSchedules from './use-job-schedules'
+
+export { useJobSchedules }
```
Mohammer5, line 3:
```suggestion
export { default as useJobSchedules } from './use-job-schedules'
```

src/hooks/job-schedules/use-job-schedules.js
```diff
@@ -0,0 +1,26 @@
+import { useDataQuery } from '@dhis2/app-runtime'
+
+const key = 'schedules'
```
Mohammer5, line 3: I think it'd be easier to read the code if this variable didn't exist.

src/hooks/job-schedules/use-job-schedules.js
```diff
@@ -0,0 +1,26 @@
...
+
+    // Remove nesting from data and move the id up
+    if (fetch.data?.[key]?.map) {
+        const data = fetch.data[key].map((job) => {
+            const id = job.sequence?.[0]?.id
+            return { ...job, id }
+        })
+
+        return { ...fetch, data }
+    }
```
Mohammer5, line 21: Just a suggestion to make it easier to read

```suggestion
    const fetch = useDataQuery(query)
    const schedules = fetch.data?.schedules

    // Remove nesting from data and move the id up
    if (schedules?.map) {
        const data = schedules.map((job) => {
            const id = job.sequence?.[0]?.id
            return { ...job, id }
        })

        return { ...fetch, data }
    }
```

src/hooks/job-types/use-job-type-parameters.js
```diff
@@ -0,0 +1,22 @@
...
+    if (fetch.data) {
+        const hasParameters = 'jobParameters' in fetch.data
+
+        if (!hasParameters) {
+            return { ...fetch, data: [] }
+        }
+
+        const data = fetch.data.jobParameters
+
+        return { ...fetch, data }
```
Mohammer5, line 16:
```suggestion
        const data = fetch.data.jobParameters
        return { ...fetch, data: data || [] }
```

src/hooks/parameter-options/index.js
```diff
@@ -0,0 +1,3 @@
+import useParameterOption from './use-parameter-option'
+
+export { useParameterOption }
```
Mohammer5, line 3:
```suggestion
export { default as useParameterOption } from './use-parameter-option'
```

src/pages/JobList/JobList.js
```diff
@@ -1,74 +1,87 @@
...
+    const [jobFilter, setJobFilter] = useJobFilter()
+    const [showSystemJobs, setShowSystemJobs] = useShowSystemJobs()
+    const { data, loading, error, refetch } = useJobSchedules()
+
+    if (loading) {
+        return <Spinner />
+    }
+
+    if (error) {
+        throw error
```
Mohammer5, line 24: Why not displaying the error like in other components?

Mohammer5 (approved):
Code looks good and apps seems to work as well. I only have a few suggestions, but I don't request any changes.

* I think it'd make sense to add a `src/hooks/index.js` so you don't have to import hooks from specific folders but just from `../../hooks/index.js`
* I've added some comments about using the `export { default as ... } from '...'` syntax, there are more files to which this comment applies, I just didn't add a comment to those files

src/pages/JobList/JobList.js
ismay: Yeah you're right. This is something I still need to address 👍

src/hooks/job-schedules/use-job-schedules.js
ismay: Yeah the reason I'm using this is because runtime requires a key to nest the data under, but I want to get rid of the nesting from the returned data. So I'm using this to ensure a clear link between the query and the destructuring. If I remove this variable I feel the link wouldn't be as clear (line 5 and 14, 15). Or is that not what you mean?

Mohammer5: What I meant is that I think it's not really that necessary to point out the link (as that's something you need to know anyway when using the app runtime), also considering [this](https://github.com/dhis2/scheduler-app/pull/495/files/107a212f5527d8b93fb2e0fb3438101ee74523d0#r1181481449) suggestion:

```jsx
const query = {
    schedules: {
        resource: 'scheduler',
    },
}

const useJobSchedules = () => {
    const fetch = useDataQuery(query)
    const schedules = fetch.data?.schedules

    // Remove nesting from data and move the id up
    if (schedules?.map) {
        const data = schedules.map((job) => {
            const id = job.sequence?.[0]?.id
            return { ...job, id }
        })

        return { ...fetch, data }
    }

    return fetch
}
```

But as I said, this is just a suggestion, feel free to leave it the way it is if you feel otherwise. It's just a tiny code snippet anyway

src/components/JobTable/Actions.js
ismay: Yeah I agree, fixed in [4fccf8b](https://github.com/dhis2/scheduler-app/pull/495/commits/4fccf8bb666d8f2432e707666923c4173c019a8b)

src/components/JobTable/Actions.js
ismay: Yeah I agree, fixed in [4fccf8b](https://github.com/dhis2/scheduler-app/pull/495/commits/4fccf8bb666d8f2432e707666923c4173c019a8b)

src/hooks/job-schedules/index.js
ismay: Addressed in [f53059b](https://github.com/dhis2/scheduler-app/pull/495/commits/f53059ba406181d3fce93f733bc6ae4249d73553)

ismay:
>     * I think it'd make sense to add a `src/hooks/index.js` so you don't have to import hooks from specific folders but just from `../../hooks/index.js`

I made this change, but it confuses the eslint import plugin. It can no longer follow the imports it seems, and we get an error about an unused module. Since this plugin is pretty useful for catching unused modules and broken imports I've reverted the change. I figured it's more important to have this properly linted than to have the benefits of the terser import style.

src/hooks/job-schedules/use-job-schedules.js
ismay: > What I meant is that I think it's not really that necessary to point out the link (as that's something you need to know anyway when using the app runtime)

True, you kind of have know about that when using the app-runtime. But I feel that making it extra clear isn't bad. Makes it very obvious when refactoring. It's a bit ugly, but I prefer ugly and clear so long as it helps others understand the code.

Also, just using the `fetch.data` in the if is on purpose. Since I modify the returned data I want to make sure that that condition applies whenever data is truthy. Just so that the hook never returns unmodified data accidentally.

ismay: I'll resolve this, but let me know if you have any remarks on the approach.

src/hooks/job-schedules/use-job-schedules.js
ismay: I modified the hooks a little in [f53059b](https://github.com/dhis2/scheduler-app/pull/495/commits/f53059ba406181d3fce93f733bc6ae4249d73553). All hooks now have a condition that applies anytime fetch.data is truthy. Inside that if I've applied your suggestion wherever possible. Let me know if there's anything that should still be addressed.

src/hooks/job-types/use-job-type-parameters.js
ismay: Fixed in [0f62c1a](https://github.com/dhis2/scheduler-app/pull/495/commits/0f62c1a4f3b8462dab031f580ec507d772950367)

src/pages/JobList/JobList.js
ismay: Fixed in [890a009](https://github.com/dhis2/scheduler-app/pull/495/commits/890a009186b7806a650f6568a21c38a9174107ca)

tomzemp (approved):
Looks good to me! I just did a high-level QA-type review (as I think that is what the second review was supposed to be), so just briefly skimming through the code changes and looking at the app. App looked like it worked well.

A couple of things that I noticed:

- the structure of useParameterOptions/useParameterOption hooks seems to result in unnecessary requests. Like if I select 'monitoring' as the job type, then the structure of the query also means that predictors, predictorGroups, dataIntegrity endpoints get requested even though it seems like for this job type, we just need the information from validationRuleGroups? Since the requests aren't very large and the app is more specialized, I think it's not going to cause issues, but it was something I noticed.
- it's cool that failed requests are minimally blocking, but it seems a bit hard to notice with the help text: 
image
Also, in some cases, like with job type push analysis, the select is disabled/non-blocking, but then it seems that one cannot create a valid job. I just wonder if it would be better with something slightly more prominent, or at least a message along the lines of "try refreshing the page" in the help text? If this is the agreed design, please disregard my comment.
- there's some warnings in the console `Data queries should always specify fields to return`. Not sure if you think it's worth it to update the query to get rid of those or not

ismay:
Thanks for the review!

> Looks good to me! I just did a high-level QA-type review (as I think that is what the second review was supposed to be), so just briefly skimming through the code changes and looking at the app.

Yeah exactly, my thinking as well.

> A couple of things that I noticed:
> 
> * the structure of useParameterOptions/useParameterOption hooks seems to result in unnecessary requests. Like if I select 'monitoring' as the job type, then the structure of the query also means that predictors, predictorGroups, dataIntegrity endpoints get requested even though it seems like for this job type, we just need the information from validationRuleGroups? Since the requests aren't very large and the app is more specialized, I think it's not going to cause issues, but it was something I noticed.
> 
> * it's cool that failed requests are minimally blocking, but it seems a bit hard to notice with the help text:
>       image
>   Also, in some cases, like with job type push analysis, the select is disabled/non-blocking, but then it seems that one cannot create a valid job. I just wonder if it would be better with something slightly more prominent, or at least a message along the lines of "try refreshing the page" in the help text? If this is the agreed design, please disregard my comment.
> 
> * there's some warnings in the console `Data queries should always specify fields to return`. Not sure if you think it's worth it to update the query to get rid of those or not

Yeah those are all good points.

- The useParameterOptions hook should probably be broken up into separate hooks. As you say, no need to request unnecessary data. I think that that will be good to tackle along with the upcoming move to a less dynamic app. Then we'll be able to match each field with its own query.
- For the fields notifying the user about errors, maybe a better solution would be to use the validationText with an error label (https://ui.dhis2.nu/demo/?path=/story/single-select-field--status-error). Though it seems slightly wrong to me semantically, it does yield a more prominent message. Or, we could also render a noticebox instead of the input. Bit jarring, but less wrong semantically and it does not seem wrong to attract the user's attention since there's an error anyway.
- And yeah, some of the queries still aren't as specific as they should be. I've noted it under the follow up points. I do think it's good to address, as well as getting rid of the unpaginated queries.

Let me know what you think.

tomzemp:
Sounds good on those points 👍

> For the fields notifying the user about errors, maybe a better solution would be to use the validationText with an error label (https://ui.dhis2.nu/demo/?path=/story/single-select-field--status-error). Though it seems slightly wrong to me semantically, it does yield a more prominent message. Or, we could also render a noticebox instead of the input. Bit jarring, but less wrong semantically and it does not seem wrong to attract the user's attention since there's an error anyway.

Yeah. I was wondering the same what would be appropriate here. I guess a [noticebox without a title](https://ui.dhis2.nu/demo/?path=/story/notice-box--without-title) would not be too intrusive? Maybe Joe would want to weigh in?

ismay:
> Yeah. I was wondering the same what would be appropriate here. I guess a [noticebox without a title](https://ui.dhis2.nu/demo/?path=/story/notice-box--without-title) would not be too intrusive?

Yeah that sounds good to me. I'm already doing that in several places, so it also fits well with the existing patterns. I'll add it to the follow ups. Once the queue functionality has landed we could ask Joe to take a look at these patterns.

PR #516, opened 2023-05-23
test: remove skipped tests

ismay:
I don't think it's a good idea to have skipped tests on master (it encourages skipping other tests as well) so I've removed them. I looked at fixing them, but the Transfer throws an error about a missing backgroundColor when mounted in enzyme, so it should probably be fixed in ui. I was hoping that a ui update would fix it, but it didn't. We looked at this in the past, I'll create an issue for this if it doesn't already exist (edit: already exists: https://dhis2.atlassian.net/browse/LIBS-342).

See also the removed comment:

```
        // At the time of writing these tests have a problem relating to the error
        // ReferenceError: backgroundColor is not defined
        // seems to be a problem in `node_modules/@dhis2-ui/transfer/build/cjs/right-side.js`
```

I've also removed the abstraction in the aggregate data component. There's no need to be very DRY in tests in my opinion. It can make them harder to follow and change.


ismay:
> Not sure which merge strategy is being used in this repo, but with these changes, I think it should be either standard or rebase merge, not a sqush merge.

You mean because there's an update and some changes to the tests in this PR? I'll clean them up a bit and then rebase merge.

PR #550, opened 2023-07-19
feat: add edit sequence route

ismay:
https://dhis2.atlassian.net/browse/DHIS2-15668

This adds the scheduler edit sequence route. For testing:

- locally the route is `http://localhost:3000/#/edit-sequence/{sequence-id}`.
- On debug 2.40 I've created a sequence and a couple jobs for testing: http://localhost:3000/#/edit-sequence/RWcaltWoKuN. 
- Since the sequence gets its id from the first job, any sequence that has job 1 as the first job will have that id.

Note that the some of the patterns used are things we might want to refactor later. I've stuck with the existing approach to maintain the consistency of the code until then.

Todo

- [x] Resize observer crash
- [x] Discard form state lifting (maybe remove discard button from outside of form)
- [x] Remove outdated comments
- [x] Check missing elements compared to edit job route (details area, etc.)
  - [x] Title to `Sequence: {name}`
  - [x] Sequence details (created, last run) (don't have that info: https://dhis2.stoplight.io/docs/dhis2/362efac6198ea-get-a-scheduler)
  - [x] Delete sequence button
- [x] Update tests
  - [x] Delete sequence modal
  - [x] Delete sequence button
  - [x] SequenceEditForm
  - [x]  SequenceEditFormContainer
- [x] Check necessity of initialValues reduce (replace with simpler approach)

Backend

- [x] https://dhis2.stoplight.io/docs/dhis2/6c6eff62bea73-delete-a-scheduler-queue, delete sequence does not work when sending a DELETE without a request body (responds with not found). app-runtime does not support a request body for DELETE requests. https://dhis2.slack.com/archives/CPGUFVC56/p1690972070298669

Follow up

- https://dhis2.atlassian.net/browse/DHIS2-15669
- https://dhis2.atlassian.net/browse/DHIS2-15670
- https://dhis2.atlassian.net/browse/DHIS2-15671
- https://dhis2.atlassian.net/browse/DHIS2-15672
- https://dhis2.atlassian.net/browse/DHIS2-15673
- https://dhis2.atlassian.net/browse/DHIS2-15677
- https://dhis2.atlassian.net/browse/DHIS2-15678


helladawit (approved):
Tested successful on version 2.41

ismay:
Also @tomzemp, if you want to test the UI:

- Go to https://pr-550--dhis2-scheduler.netlify.app/#/add-sequence/
- Set a name, a cron expression and add at least two jobs to the sequence
- Click save 
- After saving you'll be taken to the job overview
- You'll see the newly created sequence here as well
- Click actions > edit on the newly created sequence
- This will take you to the job edit screen, with a URL like https://pr-550--dhis2-scheduler.netlify.app/#/edit/xkxSSOLKc7X.
- Once there, change the `edit` in the URL to `edit-sequence` . So for example, from https://pr-550--dhis2-scheduler.netlify.app/#/edit/xkxSSOLKc7X to https://pr-550--dhis2-scheduler.netlify.app/#/edit-sequence/xkxSSOLKc7X

And then you'll have a new sequence to edit again

src/pages/SequenceEdit/SequenceEdit.js
```diff
@@ -0,0 +1,63 @@
...
+
+    if (fetching) {
+        return <Spinner />
+    }
+
+    if (error) {
+        return (
+            <NoticeBox
+                error
+                title={i18n.t('Could not load requested schedule')}
```
tomzemp, line 22: Do you want to refer to these as schedule here? Or sequence? (Also applies in next translation string)

src/components/Buttons/DeleteSequenceButton.js
```diff
@@ -0,0 +1,33 @@
+import React, { useState } from 'react'
+import PropTypes from 'prop-types'
+import { Button } from '@dhis2/ui'
+import i18n from '@dhis2/d2-i18n'
+import { DeleteSequenceModal } from '../Modal'
+
+const DeleteSequenceButton = ({ name, onSuccess }) => {
```
tomzemp, line 7: this is almost the same as the DeleteFormButton component, so I wonder if it would be useful to just reuse the existing component with some adjustments (up to you)

tomzemp (approved):
Looks good! I don't have many line-by-line comments, but a couple of things I noticed / had questions about:

- Clicking edit for a sequence from the overview scheduled jobs list sends you to the edit job page, rather than the edit-sequence page. I assume that's going to be changed later?
- the `useJobScheduleById` hook only seems to be used for the job sequences, so would it make sense to rename to `useJobSequenceById`?
- the `useJobSchedules` hook is reassigning the id of the first job to the overall schedule (I understand this is a workaround for the fact that the backend isn't creating an id for the overall sequence/schedule), but this has the effect of making schedules with only one job accessible from the edit-sequence page. Maybe if `useJobScheduleById` becomes more explicitly for sequences, it should have logic to return an error if the sequence only has one job (in which case, it wouldn't be a sequence but a job)?

ismay:
Thanks!

> * Clicking edit for a sequence from the overview scheduled jobs list sends you to the edit job page, rather than the edit-sequence page. I assume that's going to be changed later?

Definitely. When updating the list view I'll update the edit button for sequences as well.

> * the `useJobScheduleById` hook only seems to be used for the job sequences, so would it make sense to rename to `useJobSequenceById`?

Yeah the wording is a bit messed up atm. I created this issue for it: https://dhis2.atlassian.net/browse/DHIS2-15671. I could tackle that before merging.

> * the `useJobSchedules` hook is reassigning the id of the first job to the overall schedule (I understand this is a workaround for the fact that the backend isn't creating an id for the overall sequence/schedule), but this has the effect of making schedules with only one job accessible from the edit-sequence page. Maybe if `useJobScheduleById` becomes more explicitly for sequences, it should have logic to return an error if the sequence only has one job (in which case, it wouldn't be a sequence but a job)?

Yeah this is a bit confusing conceptually. So the schedules are a mix of jobs and queues. A schedule with one item in the sequence is a job, and a schedule with multiple items in the sequence is a queue. Conceptually schedules are there for both jobs and queues. The idea being that it simplifies displaying both types in the overview. I've collected this and other api quirks I've encountered so far here: https://dhis2.atlassian.net/browse/DHIS2-15669. Let me know what you think of that.

src/pages/SequenceEdit/SequenceEdit.js
ismay: Yeah I think that this should be called a queue actually. I've created a follow up issue for it.

src/components/Buttons/DeleteSequenceButton.js
ismay: Yeah. Currently I think we have the delete sequence/job buttons in the list view, and the delete buttons in the edit views. The modal is at the bottom of the hierarchy and includes the hook for the request.

We could have a button that shows a component on click, with the component being a modal. That seems to be the generic bit of functionality here. I'll create a follow up issue for it.

ismay: https://dhis2.atlassian.net/browse/DHIS2-15677

PR #559, opened 2023-08-15
fix: add new route structure

ismay:


src/components/Routes/Routes.js
```diff
@@ -1,22 +1,25 @@
...
-        <Route exact path="/" component={JobList} />
-        <Route path="/edit/:id" component={JobEdit} />
-        <Route path="/view/:id" component={JobView} />
-        <Route path="/add" component={JobAdd} />
-        <Route path="/add-sequence" component={SequenceAdd} />
-        <Route path="/edit-sequence/:id" component={SequenceEdit} />
+        <Switch>
+            <Redirect from="/edit/:id" to="/job/:id" />
+            <Redirect from="/view/:id" to="/job/:id" />
+            <Redirect from="/add" to="/job/add" />
```
Mohammer5, line 16: What about `/add-sequence -> /queue/add` and `/edit-sequence/:id -> /queue/:id`?

Mohammer5 (changes requested):
Found only one thing

src/components/Routes/Routes.js
ismay: Those routes are not referenced anywhere yet, so noone should be using them. On a related note, the scheduler -> queue renaming thing I'm going to address in this issue: https://dhis2.atlassian.net/browse/DHIS2-15671

PR #560, opened 2023-08-16
refactor: improve hooks and naming

ismay:
- https://dhis2.atlassian.net/browse/DHIS2-15671
- https://dhis2.atlassian.net/browse/DHIS2-15678

Todo

- [x] use-job-by-id should just fetch single job
- [x] Load queue from queues endpoint, reference by name
- [x] Update tests
- [x] Remove use-schedule-by-id
- [x] Distinguish between schedule, queue and sequence
- [x] Ensure that queue names are uri encoded when used in the url

Follow up

- There are some areas that still need to be updated to the new names. I thought I'd leave that update to when we update the list view (after this task), as it's all somewhat tied to that task. See the integration tests, the joblist route and some of the components.

Concepts

There are several concepts used throughout the app that have to do with the queueing functionality. The usage was a little sloppy, so I've started cleaning it up in this PR:

- A `job` is a single task that can be planned with a cron expression, or to run with a certain interval between executions (delay).
- A `queue` or `job queue` is a collection of jobs that will be executed in sequence.
- A `queueable` is a job that is not part of a queue (not used in the ui)
- We retrieve items from the `/scheduler` endpoint that are meant to simplify rendering the root list view. These can be either jobs or queues. Instead of calling these `schedules` after the endpoint, we're calling these `jobsAndQueues`. This is to prevent confusion with the `schedule` property of jobs and queues.

In the ui we're using the concepts `job queue` and `job` (so the user is only aware of those terms). We also use the heading `Scheduled jobs` in the list view, since the queues and jobs essentially are all scheduled jobs.


src/components/Forms/QueueAddFormContainer.js
```diff
@@ -0,0 +1,24 @@
...
+import QueueAddForm from './QueueAddForm'
+
+const { Form } = ReactFinalForm
+
+const QueueAddFormContainer = () => {
+    const redirect = () => {
+        history.push('/')
+    }
+    const [submitQueue] = useSubmitQueue({ onSuccess: redirect })
+
```
Mohammer5, line 14:
```suggestion

const navigateHome = () => history.push('/')

const QueueAddFormContainer = () => {
    const [submitQueue] = useSubmitQueue({ onSuccess: navigateHome })

```

Very, very, very minor comment: `navigateHome` seems more descriptive to me (`onSuccess: navigateHome`). And, as `history` is being imported directly, the function can be outside of the render code.

src/components/Forms/QueueEditFormContainer.js
```diff
@@ -0,0 +1,59 @@
...
+
+const QueueEditFormContainer = ({ queue, jobs }) => {
+    const redirect = () => {
+        history.push('/')
+    }
+
+    const [submitQueue] = useUpdateQueue({
+        onSuccess: redirect,
+        initialName: queue.name,
+    })
```
Mohammer5, line 18: Same as above

src/pages/QueueEdit/QueueEdit.js
```diff
@@ -3,38 +3,35 @@ import { Card, IconInfo16, NoticeBox } from '@dhis2/ui'
...
+import { useJobs } from '../../hooks/jobs'
+import styles from './QueueEdit.module.css'
 
-const SequenceEdit = () => {
-    const { id } = useParams()
-    const { data, fetching, error } = useJobScheduleById(id)
+const QueueEdit = () => {
+    const { name } = useParams()
+    const queueFetch = useQueueByName(name)
+    const jobsFetch = useJobs()
```
Mohammer5, line 14: Not a fan of the term `fetch` here. Without context, I'd expect this to be a `fetch`-function with specific parameters/headers. react-query uses the term `result`, so maybe `queueResult` and `jobsResult`?

Mohammer5 (approved):
Looks good! Found a few very, very minor things. Up to you whether and how you want to address them

src/pages/QueueEdit/QueueEdit.js
ismay: Yeah, that makes sense. I've updated it.

src/components/Forms/QueueAddFormContainer.js
ismay: Yeah you're right, that's clearer. Updated it.

ismay:
> I added a queue (directly from the URL) and that replaces the individual jobs in the list .. I guess that's just what you get back from the API, right? it's a bit confusing UX-wise
> the queue also doesn't have a "type" in the list, but I take it that will be coming in a different PR?

It is yes. The list view currently shows queues, but that's just a side effect of switching to the `scheduler` endpoint for the data we render there. The list view is the last major task, we'll properly update the list view to show queues and the jobs included in queues with that issue. The logic that's being used there now only renders jobs properly, so I agree that it's a bit confusing atm.

ismay:
> This summary you have in the PR description for what's a job vs queue vs queuable .. it would be good to add to the README. It's not very clear and something that would help me, if I am to pick up a task in the app, to have some extra context.

Ah yeah good point. I'll add it.

PR #571, opened 2023-10-17
feat: incorporate job queue in job list

ismay:
closes #575
closes #576 
closes #577
closes #578

https://dhis2.atlassian.net/browse/DHIS2-14749

- https://www.figma.com/file/YVXIsd80IgZnTfMDE0qOGf/Scheduler?type=design&node-id=145-1279&mode=design
- https://dhis2.stoplight.io/docs/dhis2/ehaxt2lzojavb-dhis-2-api-scheduler-only
- https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-master/scheduling.html

Todo:

- [x] Add "new sequence" button
- [x] Update table headers to match design
- [x] Add collapsible rows for queues
- [x] Update job toggle to json patch
- [x] Finish naming updates from: https://github.com/dhis2/scheduler-app/pull/560, rename all concepts to new terms, see readme.
- [x] Update/add tests
- [x] Update outdated documentation links
- [x] See if queued job rows should be colored differently, or maybe indented, so they're easy to distinguish
- [x] Maybe animate showing and hiding of rows
- [x] Align dropdown icon appearance with other icons
- [x] Get rid of bottom border on expanded row
- [x] Replace inline button styles with external css
- [x] Update job toggle to new endpoint
- [x] e2e: filter, toggle, actions

Follow up:

- Jobs that are part of a queue are no longer editable through the UI, would be nice to have
- We could support system queues in the future
- Sort jobs/queues in list
- Status label doesn't update correctly when toggling a job
- Allow running queues manually
- Switch next run time label and content. The way we're doing it now is contrary to the way it's commonly done. We're displaying the complex, hard to understand information first, and the easier to understand at a glance information second. Someone who understands the former doesn't really need the latter.
- A job without a schedule can be enabled
- The relation between jobs and queues is not transparent to the user and not very convenient to work with for the frontend. For example, the status of a queue depends on the activation status of the first job. If you move a job that was disabled before it was added to the queue, to the first position, the queue will be disabled. This is not transparent to the user. It would make more sense for the model to align with what we present to the user: a queue being its own entity. Where the activation of a queue and its schedule is independent of the underlying job. This could then also hopefully mean that the schedule of jobs does not have to be erased when they're moved into a queue, allowing it to be preserved for when its moved out of the queue. It would also get rid of the akward dynamic between the id of the first job, and the name of a queue, where sometimes when dealing with queues you need the id of the first job (enabling/disabling a queue), and sometimes you need the name of the queue (editing a queue).
- Add `cursor: pointer` to transfer options
- We could support delay queues

Backend:

- ~~Deleting a queue does not result in the jobs being available from the scheduler endpoint again, but the names of the jobs are still reserved, so it seems they're still present somehow. (Fixed and available on debug dev)~~
- ~~Toggling a queue on or off ejects the first job from the queue (Patch bug, new endpoint available now)~~


Mohammer5 (approved):
Works

tomzemp (approved):
Looks great 🙌! (FYI: I've done a QA review, so I haven't looked at the code)

I agree with the points in your notes. Particularly:

- I find the deletion (/removal of jobs) slightly confusing and how it adds back the jobs without the original scheduling
- I always got queues created in disabled mode, and that didn't really make sense to me as a user (who hasn't read the documentation for job scheduling 🤡).


**One thing I would suggest changing before merging**

The language in the cancel editing dialog is a bit confusing:
image
When I first opened the app, my UI language was in French, and the French was (slightly) mistranslated such that it was suggesting that you were deleting the form (I guess the translation memory is picking up a translation from some other app where this would be the correct meaning). 

Either way, I think the main dialog might be clearer if it said something like `Do you want to discard these changes?` (i.e. not discard the form). Also it's a bit confusing to me with the changed-my-mind button for this dialog being labeled `Cancel`, since you get to the dialog by clicking `Cancel` (so maybe you would click `Cancel` again in the dialog, thinking you were confirming your choice). Maybe the `Cancel` button within the dialog could be `Keep editing` or something similar?
<br/>

**Other things I noticed (only the first point is related to this PR)**

- I find it slightly confusing that the job type is `Queue` for queues but the specific type for job (like maybe it should be `Job: {{type}}`?
- some of the translations are missing in. the app. The ones I noticed were already missing (job types (from backend) and the cron labels), so I don't think this a new issue, but because of that I didn't review new translation coverage.
- I noticed the backend wouldn't let me save two jobs of the same type/cron expression even if the details are different (specifically I got this when creating jobs to run different data integrity checks). Not sure if you know if that's a known issue.
- It seemed that the backend stopped me from saving jobs with very long names (but then just returned generic 500 error). I guess it would be good to figure out if there's a limit and then add a validation on the input?

PR #575, opened 2023-11-24
chore: update doc link

dayures:


ismay:
Hi @dayures, thank you for the PRs. I agree that these links should be updated. This change has already been incorporated in https://github.com/dhis2/scheduler-app/pull/571, so the links should be good as soon as that's merged.

dayures:
Thanks for the feedback! Please feel free to close them as soon as the #571 is merged

kabaros (approved):

PR #598, opened 2024-02-27
fix: add html push analytics translation

ismay:


kabaros (approved):
tested this and it looks fine - the job itself fails, but I guess that has nothing to do with scheduler itself.

PR #603, opened 2024-03-11
fix(data-integrity): change default report type to summary

Birkbjo:
Changes the defaultValue of `Report` type for `Data integrity` job to be `summary` instead of `report`.
Report is a legacy type, and isn't really useful at all when scheduling. So we may remove this completely.

Also fixes a bug where I'd used `initialValue` for the field instead of `defaultValue`. This is the wrong prop to use it in this case. Because `initialValue` would override the `initialValues` on the form, and meant this field would always be set to `report` when you edited a job. Which is actually a quite bad bug.
I just discovered this bug when trying to schedule summary jobs in conjunction with the new Data Integrity UI in `Data Administration` app.
However, this bug does not seem to happen in 2.39 and 2.38. I am not sure what introduced it, but it seems like it's currectly in 2.40 and master/dev.

Also updated `Run all available checks`-label to `Run all standard checks`. Because `slow: true`-checks will not be run this is selected, so right now it's quite misleading. `Standard` is the term used in `Data Administration` UI for checks that are not slow.

### DIFF
--- a/cypress/integration/add-job/create-parameterless-jobs/index.js
+++ b/cypress/integration/add-job/create-parameterless-jobs/index.js
@@ -57,7 +57,7 @@ Then(
         saveAndExpect({
             cronExpression: '0 0 * ? * *',
             jobParameters: {
-                type: 'REPORT',
+                type: 'SUMMARY',
             },
             jobType: 'DATA_INTEGRITY',
             name: 'Name',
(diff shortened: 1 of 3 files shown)


Birkbjo:
> Thanks, yeah that looks good to me. I'm not seeing this change btw.:
> 
> > Also updated Run all available checks-label to Run all standard checks. Because slow: true-checks will not be run this is selected, so right now it's quite misleading. Standard is the term used in Data Administration UI for checks that are not slow.

Ah, sorry that was part of the other PR. Must've confused these...

PR #604, opened 2024-03-11
fix(data-integrity): add indicator for slow checks

Birkbjo:
Adds an indicator to checks that have `isSlow: true`. The new UI in `Data Administration` separates these checks completely, because the user should be cautious when running these checks. So I think it's quite important to distinguish those here as well.

### DIFF
--- a/src/components/FormFields/Custom/DataIntegrityChecksField.module.css
+++ b/src/components/FormFields/Custom/DataIntegrityChecksField.module.css
@@ -1,21 +1,36 @@
 .transfer {
-    margin-top: 8px;
+    margin-block-start: var(--spacers-dp8);
 }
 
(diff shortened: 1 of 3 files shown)


src/components/FormFields/Custom/DataIntegrityChecksField.module.css
```diff
@@ -1,23 +1,38 @@
 .transfer {
-    margin-top: 8px;
+    margin-block-start: var(--spacers-dp8);
 }
 
 .transferOption {
+    display: flex;
```
Birkbjo, line 6: I changed this to flex, mostly because I prefer using `gap` instead of `margins` for spacing between the header and other stuff.

PR #605, opened 2024-03-12
fix: add missing parameter fields

ghost:
https://dhis2.atlassian.net/browse/DHIS2-17024

- [x] Analytics table: skipPrograms (java.util.Set)
- [x] Test:
  - [x] waitMillis
  - [x] itemDuration
  - [x] failWithPolicy
- [x] Add missing skip table types translations

Follow up:

The requirements for the tracker search optimization job's attributes field are unclear, so that field is still missing. I've asked the tracker team. We can implement it later if necessary.

Some details on the error: relativeApiEndpoint: "/api/trackedEntityAttributes/indexable" for tracker search optimization yields: "TrackedEntityAttribute with id indexable could not be found." (created here: https://dhis2.atlassian.net/browse/DHIS2-17093)


cypress.json
```diff
@@ -6,7 +6,7 @@
     "video": false,
     "env": {
         "dhis2DataTestPrefix": "dhis2-scheduler",
-        "dhis2ApiVersion": "41",
+        "dhis2ApiVersion": "42",
```
Birkbjo, line 9: This needs a `BREAKING CHANGE`-commit right? Or would you squash-merge and add that in the message? https://github.com/dhis2/scheduler-app/pull/605/commits/6756afaa2c82d1acc19b2ab55bd9b7ef71416e0d doesn't seem to contain this.

Birkbjo (changes requested):
The code looks good to me!

~I'm a bit confused about the change to use `42` though. If this is not supported in `41`, I don't think we should release this in the app in master at all for quite some time.~
(Sorry, just got a bit confused - see that this is only for the tests against dev).

Birkbjo (approved):

PR #609, opened 2024-04-09
refactor: use dockerized backend

ismay:
https://dhis2.atlassian.net/browse/LIBS-597

This replaces the backend we use for our end to end tests. Instead of mocking the backend, this runs the e2e tests against a dockerized backend. Recording fixtures is tricky to get right and a recurring, time consuming task. Besides that there was a potential for divergence between the actual backend and the mocked backend, decreasing the realism of the tests.

I've updated cypress to the latest version so that we can use their [session command](https://docs.cypress.io/api/commands/session) for a consistent browser context between tests. Because of that I've also updated @testing-library/cypress, @testing-library/react-hooks, start-server-and-test and node versions. One of our upstream libs has changed maintainers to https://github.com/badeball/cypress-cucumber-preprocessor, so I've updated that as well.

Because of these changes I've diverged from our own cypress libs:

- @dhis2/cli-utils-cypress
- @dhis2/cypress-commands
- @dhis2/cypress-plugins

I've dropped them for now. We could integrate these changes upstream, though I do think there's also a potential for dropping some of the dhis2 cypress libraries and decreasing our maintenance burden. We'll have to see where the centralization is working to our benefit, and where it isn't. I think that if we dropped cucumber there would be very little reason to have our own cypress libs, with the exception of maybe the login helper.

Follow up:

- Potential upstreaming of these changes (or dropping library functionality)
- Ensure docker images are safe to rely on (channel dev?)
- Improve running time (building docker images with seeded db built in makes a difference/parallelization as well)
- Potentially run e2e tests on a timer
- Final form registering as dirty immediately after render

closes #484
closes #489


cypress/fixtures/edit-job/single-user-job.json
```diff
@@ -9,7 +9,7 @@
     "enabled": true,
     "leaderOnlyJob": false,
     "externalAccess": false,
-    "jobType": "DATA_INTEGRITY",
+    "jobType": "DISABLE_INACTIVE_USERS",
```
ismay, line 12: I've changed this because the data integrity job registers as modified after loading due to a final form quirk. This caused the test to fail due to a popup, something that didn't happen with the mocked backend (even though it should have). I'll create a follow up issue for addressing the final form quirk.

cypress.config.js
```diff
@@ -0,0 +1,48 @@
+// eslint-disable-next-line import/no-unused-modules
+const { defineConfig } = require('cypress')
+const webpack = require('@cypress/webpack-preprocessor')
+const preprocessor = require('@badeball/cypress-cucumber-preprocessor')
+
+/**
+ * Configuration for the cypress-cucumber preprocessing, see:
+ * https://github.com/badeball/cypress-cucumber-preprocessor/tree/master/examples/webpack-cjs
+ */
+async function setupNodeEvents(on, config) {
```
ismay, line 10: This could be dropped if we drop cucumber.

Mohammer5 (approved):
LGTM

kabaros:
Yeah, will do after the academy .. thanks @ismay

PR #618, opened 2024-06-10
ci: add reusable workflows

ghost:
See: https://dhis2.atlassian.net/browse/CLI-71

Docs / links

- https://docs.github.com/en/actions/using-workflows/avoiding-duplication
- https://docs.github.com/en/actions/using-workflows/reusing-workflows
- https://github.com/dhis2/workflows/
- https://github.com/cypress-io/github-action/blob/master/.github/workflows/main.yml
- https://github.com/cycjimmy/semantic-release-action
- https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflows-in-forked-repositories
- https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks
- https://docs.github.com/en/actions/using-workflows/creating-starter-workflows-for-your-organization
- https://github.com/dhis2/workflows-test
- https://resources.github.com/learn/pathways/automation/intermediate/create-reusable-workflows-in-github-actions/

Todo

- [x] Cancel previous workflows only on non release branches
- [x] Replace comment and close wf with approach referenced here: https://github.com/peter-evans/close-issue
- [x] https://docs.github.com/en/actions/using-workflows/using-github-cli-in-workflows
- [x] See where inputs might be needed (see other apps, libs)
- [x] Abstract to other repo
- [x] Verify if expressions
- [x] Use consistent order for contains expressions
- [x] Check concurrency groups
- [x] Update dependencies
- [x] Set up publishing workflow (and refer to versioned workflows)
- [x] Investigate how to migrate from old strategy
- [x] Find alternative to JulienKode/pull-request-name-linter-action

Notes

- Migration:
  - Users currently install the latest workflows from [here](https://github.com/dhis2/workflows/tree/master/ci) with cli-style. The new flow would be using [github starter workflows](https://docs.github.com/en/actions/using-workflows/creating-starter-workflows-for-your-organization), and refer to workflows defined [here](https://github.com/dhis2/workflows-platform/tree/main/.github/workflows). There is no clash between the old way of working and the new one, so we can migrate gradually.
- Discussion in this issue elaborates a little on the release process for reusable workflows that we're using: https://github.com/semantic-release/semantic-release/issues/1515 and https://github.com/doteric/semantic-release-major-tag. See this commit: https://github.com/dhis2/workflows-test/commit/26db1477d9b38ce7f2351bf0cfacaff25051a2a0

Follow up

- Update commitlint workflow to v6, requires an update to cli-style
- Add starter workflows
- Incorporate workflows from other apps/libs/repos
- Move to renovate, also for wf repo to keep actions up to date. See if renovate can also upgrade reusable workflows
- Lots of warnings for dhis2/action-semantic-release@master, dhis2/deploy-build@master. Needs to be updated or they will fail soon

### DIFF
--- a/.github/workflows/deploy-pr.yml
+++ b/.github/workflows/deploy-pr.yml
@@ -4,7 +4,7 @@ on:
     pull_request:
 
 concurrency:
-    group: ${{ github.workflow}}-${{ github.head_ref }}
+    group: ${{ github.workflow }}-${{ github.head_ref }}
     cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/deploy-production.yml
+++ b/.github/workflows/deploy-production.yml
@@ -6,7 +6,7 @@ on:
             - master
 
 concurrency:
-    group: ${{ github.workflow}}-${{ github.ref }}
+    group: ${{ github.workflow }}-${{ github.ref }}
     cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/test-and-release.yml
+++ b/.github/workflows/test-and-release.yml
@@ -3,7 +3,7 @@ name: test-and-release
 on: push
 
 concurrency:
-    group: ${{ github.workflow}}-${{ github.ref }}
+    group: ${{ github.workflow }}-${{ github.ref }}
     # Cancel previous runs if not on a release branch
     cancel-in-progress: ${{ !contains(github.ref, fromJSON('["master", "main"]')) }}
 
@@ -20,7 +20,7 @@ jobs:
         if: '!github.event.push.repository.fork'
         secrets: inherit
     release:
-        needs: [lint, test, e2e]
+        needs: [lint-commits, lint, test, e2e]
         uses: dhis2/workflows-platform/.github/workflows/release.yml@v1
         # Skips forks and dependabot PRs
         if: '!github.event.push.repository.fork'
--- a/.github/workflows/lint-pr-title.yml
+++ b/.github/workflows/lint-pr-title.yml
@@ -5,7 +5,7 @@ on:
         types: ['opened', 'edited', 'reopened', 'synchronize']
 
 concurrency:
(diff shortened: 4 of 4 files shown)


.github/workflows/deploy-pr.yml
```diff
@@ -0,0 +1,14 @@
+name: deploy-pr
+
+on:
+    pull_request:
+
+concurrency:
+    group: ${{ github.workflow}}-${{ github.head_ref }}
```
Mohammer5, line 7:
```suggestion
    group: ${{ github.workflow }}-${{ github.head_ref }}
```

.github/workflows/deploy-production.yml
```diff
@@ -0,0 +1,17 @@
+name: deploy-production
+
+on:
+    push:
+        branches:
+            - master
+
+concurrency:
+    group: ${{ github.workflow}}-${{ github.ref }}
```
Mohammer5, line 9:
```suggestion
    group: ${{ github.workflow }}-${{ github.ref }}
```

.github/workflows/test-and-release.yml
```diff
@@ -0,0 +1,27 @@
...
+        uses: dhis2/workflows-platform/.github/workflows/lint.yml@v1
+    test:
+        uses: dhis2/workflows-platform/.github/workflows/test.yml@v1
+    e2e:
+        uses: dhis2/workflows-platform/.github/workflows/e2e.yml@v1
+        # Skips forks and dependabot PRs
+        if: '!github.event.push.repository.fork'
+        secrets: inherit
+    release:
+        needs: [lint, test, e2e]
```
Mohammer5, line 23: I know nothing has changed, but why does this depend on those 3 jobs, but not on lint-commits?

Mohammer5 (approved):
LGTM, not sure why there are some actions not being run that block you from merging?

PR #644, opened 2025-01-20
fix: cron parsing number range [DHIS2-18859]

tomzemp:
See https://dhis2.atlassian.net/browse/DHIS2-18859

This fixes an issue where the parsing of cron expressions would compare two strings, e.g. `'4'`, and `'23'` and '4' < '23' evaluates to `false`.

### DIFF
--- a/src/services/validators/validate-cron.js
+++ b/src/services/validators/validate-cron.js
@@ -28,7 +28,7 @@ const isValidNumberRange = (range, x, y) => {
     return (
         isValidNumber(boundaries[0], x, y) &&
         isValidNumber(boundaries[1], x, y) &&
-        boundaries[0] <= boundaries[1]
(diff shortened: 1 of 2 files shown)


src/services/validators/validate-cron.js
```diff
@@ -28,7 +28,7 @@ const isValidNumberRange = (range, x, y) => {
     return (
         isValidNumber(boundaries[0], x, y) &&
         isValidNumber(boundaries[1], x, y) &&
-        boundaries[0] <= boundaries[1]
+        Number(boundaries[0]) <= Number(boundaries[1])
```
tomzemp, line 31: since these have already been evaluated as valid numbers, we should be able to use `Number()` here

Chisomchima (approved):
lgtm!

